Issue #6, opened 2020-12-17
feat: add CSS styles for buttons


lauthieb:
**Feature request**

Add CSS for buttons specified in our documentation: https://www.decathlon.design/726f8c765/p/8008f8-button/b/50afe1

Issue #3, opened 2020-12-17
feat: add CSS design tokens (colors, typography, spacings, breakpoints)


lauthieb:
**Feature request**

On [decathlon.design](https://decathlon.design), we already have several design tokens to add in our CSS package (#1).

**Preferred solution**

Use CSS Custom Properties (CSS variables)

**Alternatives**

Feel free to comment below.

**Additional context**

Styles are already available in our documentation:
- [Colors](https://www.decathlon.design/726f8c765/p/85c5a1-colors/b/06f55e)
- [Typography](https://www.decathlon.design/726f8c765/p/860e14-typography/b/208916)
- [Spacings & Breakpoints](https://www.decathlon.design/726f8c765/p/061870-layout/b/91b22f)

Issue #1, opened 2020-12-17
config: add package for CSS styles


lauthieb:
**Feature request**

In order to separate styles from markup and interaction, we have to create a package dedicated only to CSS styles (design tokens + components classes). This will be used as a dependency of our next component libraries (Web Components, React, Svelte...).

PR #16, opened 2020-12-18
chore(init): multiple packages architecture with Lerna & CSS package with PostCSS + Tailwind CSS + Storybook

lauthieb:
## Pull Request checklist
🚨Please review the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch**. Don't request your master!
- [x] Make sure you are making a pull request against the **master branch** (left side).

## Pull request type

- Config

Issues Numbers: #17 #1 #3 #6

- No

## TODO

- [x] init Lerna 
- [x] init CSS package with Tailwind CSS
- [x] `error postcss@8.2.1: The engine "node" is incompatible with this module. Expected version "^10 || ^12 || >=14". Got "13.9.0"`
- [x] stylelint
- [x] commitlint
- [x] husky configuration in separated JSON file
- [x] README documentation (install yarn)
- [x] engines in `package.json` (force to use yarn instead of NPM)
- [x] reorder properties in `package.json` files
- [x] `lerna Unknown argument` (`lint-staged` + `yarn run eslint` bug)

❤️ Thank you!


lauthieb:
@jlind23 @jleveugle @gabbloquet, I'm in draft phase but feel free to begin the review :-)

packages/css/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@decathlon/vitamin-css",
+  "version": "0.0.1",
+  "description": "CSS styles for Vitamin Web",
+  "author": "Decathlon <designsystem@decathlo>",
+  "homepage": "https://github.com/Decathlon/vitamin-web#readme",
+  "scripts": {
+    "build": "cross-env NODE_ENV=production postcss src/index.css -o lib/index.css",
+    "storybook": "npm run build && start-storybook -s ./lib -p 6006",
```
amen-souissi, line 9: I think your system does not support watch mode. I suggest you split your scripts like this:
{
...
"build": "cross-env NODE_ENV=production npm run build:tailwind",
"build:tailwind": "postcss src/index.css -o lib/index.css",
"watch:tailwind": "chokidar 'src/**/*.css' -c 'npm run build:tailwind'",
 "storybook": "npm-run-all build:tailwind --parallel watch:tailwind start:storybook",
"start:storybook": "start-storybook -p 6006 -s public",
...  
}

amen-souissi (changes requested):
GG :), small changes

packages/css/tailwind.config.js
```diff
@@ -0,0 +1,72 @@
...
+  theme: {
+    screens: {
+      mobile: '0px',
+      tablet: '600px',
+      'small-desktop': '896px',
+      'medium-desktop': '1200px',
+      'large-desktop': '1800px',
+    },
+    colors: {
+      black: '#001018',
```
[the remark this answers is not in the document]
lauthieb: Yes, this idea could be nice but because it's not documented by designers, I don't think developers will use this notation. For example, on Figma file, it will be written "brand-digital-light-2" and no "100".

packages/css/tailwind.config.js
```diff
@@ -0,0 +1,72 @@
+module.exports = {
+  prefix: 'vtmn-',
+  purge: ['src/stories/**/*.mdx'],
+  darkMode: false, // or 'media' or 'class'
+  theme: {
+    screens: {
```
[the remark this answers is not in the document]
lauthieb: Good idea!
Done: https://github.com/Decathlon/vitamin-web/pull/16/commits/59fbc23c4fee65f6e769a833244825ad8c1f6dcd

packages/css/tailwind.config.js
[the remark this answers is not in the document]
amen-souissi: brand-digital: {
dark-1: 'var(--vtmn-color_brand-digital-dark-1)',
dark-2: 'var(--vtmn-color_brand-digital-dark-2)',
...
},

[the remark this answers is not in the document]
lauthieb: Done :)

lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "packages": [
+    "packages/*"
+  ],
+  "version": "0.0.1"
```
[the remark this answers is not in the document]
thollander: It really depends on what's going to be aside to `packages/css`.
What I understand is that there'll be a `packages/react` next to it, consuming `packages/css`. 
If so, my opinion is that it would be interesting to let the default lerna setting which is "fixed".

As a react consumer, I wouldn't want to have to check the internal version of the `css` package to know the used version as it can differ from the other one. (not the same lifecycle...) 

If most of the time, all packages change it would be better (IMHO) to make them evolve with the same version number even if sometimes it involves blank release on some packages.

Ps : that's what Babel does and it's working well 

(Also, as there is only one package for the moment we can leave it like this and decide when another package will be added)

packages/css/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@decathlon/vitamin-css",
+  "version": "0.0.1",
+  "description": "CSS styles for Vitamin Web",
+  "author": "Decathlon <designsystem@decathlo>",
```
thollander, line 5:
```suggestion
  "author": "Decathlon <designsystem@decathlon.com>",
```

packages/css/src/components/buttons.css
```diff
@@ -0,0 +1,138 @@
+@layer components {
+  /* Default button */
+  .vtmn-btn {
+    --button-font-family: var(--vtmn-typo_font-family--condensed), Arial,
+      Helvetica, sans-serif;
+    --button-font-size: 16px;
```
thollander, line 6: Do you consider working with `em` or `rem` in the future?

packages/css/src/components/buttons.css
```diff
@@ -0,0 +1,138 @@
...
+  .vtmn-btn:not([disabled]):hover {
+    cursor: pointer;
+    background-color: var(--button-background-color-hover);
+  }
+  .vtmn-btn:not([disabled]):active {
+    background-color: var(--button-background-color-active);
+  }
+  .vtmn-btn:not([disabled]):focus {
+    outline: none;
+    box-shadow: inset 0px 0px 0px 2px var(--button-border-color),
```
thollander, line 54:
```suggestion
    box-shadow: inset 0 0 0 2px var(--button-border-color),
```
Just a small remark : `px` information here is quite useless. Making only `0` instead of `0px` will decrease the css payload

packages/css/src/components/buttons.css
gabbloquet: We are working with that because designers elements are defined like that. 
You think it could be interesting to declare it for the whole application ?

packages/css/package.json
lauthieb: Thanks, Terence!
I cannot accept your commit suggestion because it's `@decathlon.net`.
I've just made the fix, thanks a lot!

packages/css/src/components/buttons.css
lauthieb: Oh yes sure! Fixed.

packages/css/src/components/buttons.css
thollander: I guess it should be defined as some global level and defined in em unit. That's a very good practice that we should not avoid even if designers work in px (and that also a lot of design system work like this too...).
As a concrete example, while zooming on the e-commerce website, that would imply that everything adapts except buttons coming from design system 😕

Check out this interesting article https://medium.com/codyhouse/create-your-design-system-part-1-typography-7c630d9092bd

lauthieb: Thanks @thollander for this suggestion.
Are you sure that zoom is always a problem for modern browsers?
But I understand why it's important to use relative units like rem,em instead of absolute like px,cm etc.
I will this from Tailwind: https://tailwindcss.com/docs/font-size#font-sizes

lerna.json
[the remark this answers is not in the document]
lauthieb: Completely agree @thollander.
Yes, after there will probably have `packages/react`, `packages/svelte`, `packages/wd` that will consume `packages/css` styles.
Except if there are some other suggestions inside our discussion about architecture in this repository.

packages/css/tailwind.config.js
```diff
@@ -0,0 +1,85 @@
+module.exports = {
+  prefix: 'vtmn-',
```
lauthieb, line 2: @amen-souissi @gabbloquet @thollander @jleveugle what's your opinion about this prefix, please?

IMHO,
**Pros:**
- It ensures that every project that already uses Tailwind will not have conflicts.

**Cons:**
- It will make HTML & CSS more verbose (not good for performance in a big app)
- We will always have to tell "don't forget to add our prefix to utility classes, example: `vtmn-text-center` instead of `text-center`" when we will redirect the developer to Tailwind CSS documentation inside our Storybook.

Personally, now I think it will be better without prefix.

packages/css/src/components/buttons.css
lauthieb: I've just made an issue for this refactoring: #19

packages/css/package.json
lauthieb: @gabbloquet @jleveugle @thollander if you have an idea... (: 
thx!

packages/css/src/components/buttons.css
jleveugle: Hey,

We should use rem or em, as Terence said.

It's a best practice to use the value defined in the browser (by default, 16px). 

W3C recommandation:
 https://www.w3.org/TR/WCAG20-TECHS/C14.html

packages/css/tailwind.config.js
jleveugle: Hey, I think that to facilitate adoption, we should prefix because it is a safe bet that Vitamin will not be the only one present in projects in transition.

For me, the argument about performances here is negligible, because I think that we have a lot of optimisation before that, like using only the design system (and remove others CSS), purge our CSS / JS etc.

I think that we can manage this purpose layer. Are you agreed ? cc @thollander

lauthieb: @jleveugle thanks for your answer.
And what about the other problem in our documentation? Developers that use Vitamin will have to "take" the good documentation from Tailwind CSS without forget to add `vtmn-` before each utility classes.

NicolasBreunin: i agree with @jleveugle , it is possible that Vitamin is not the only design system used. to avoid any conflict, we should leave the prefix

amen-souissi: @jleveugle I agree with you. It is also for give an identity.
Another problem is that Vscode extension does not generate suggestions for Tailwind prefixes (hope this will be resolved soon ...)

lerna.json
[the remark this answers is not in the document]
jleveugle: agree with @thollander.

In any case, it will be easier to switch from "fixed" mode to "independent" mode if you realize that the last one is more appropriate.

packages/css/package.json
amen-souissi: I think it's because the css file is considered a static file. I think it must be imported into a decorator (global decorator, see my previous comment) to activate the hot reload...

packages/css/tailwind.config.js
lauthieb: Thanks for your opinion! So we will keep this prefix.

lauthieb:
@crotru it would be nice to have your opinion about this Pull Request :)

packages/css/package.json
amen-souissi: waiting for #20... thanks for @jleveugle:
using Postcss 7 instead of Postcss 8 (more stability for the watch mode)
yarn instead of npm (I think we need to discuss this)

packages/css/src/components/buttons.css
```diff
@@ -0,0 +1,138 @@
+@layer components {
+  /* Default button */
+  .vtmn-btn {
```
amen-souissi, line 3: we will have to migrate to TailwindCSS in other PR (@apply for custom css)

amen-souissi (approved):
👍

jleveugle (approved):

NicolasBreunin (approved):

crotru (approved):

packages/css/src/components/buttons.css
lauthieb: Thanks! Issue created #24

crotru:
Approved for merge but still created some side issues 🙂 :
 - #29 refactor: line-height shouldn't wear unit
 - #30 Design evolution: forbid negative line-height
 - #31 Design evolution: spacing as multiple of 16

PR #20, opened 2020-12-22
build: watch css files using Storybook

jleveugle:
Hello,

Some update for your PR @lauthieb.

Files are now watch correctly, using webpack configuration instead of statitc files.

I switch from NPM to Yarn, because this last one has more feature to manage a monorepo (like resolutions, etc.), and dependencies management is better than NPM < 7.

To be compliant with lerna, I purpose to use Yarn "Classic" (v1), but we should consider using Yarn Berry (v2) (and remove lerna, I think).

See ya !


package.json
```diff
@@ -23,52 +39,33 @@
...
     "eslint-config-airbnb": "^18.2.1",
     "husky": "^4.3.6",
     "lerna": "^3.22.1",
-    "link-parent-bin": "^2.0.0",
     "lint-staged": "^10.5.3",
     "npm-run-all": "^4.1.5",
-    "postcss": "^8.2.1",
-    "postcss-cli": "^8.3.1",
-    "postcss-import": "^14.0.0",
+    "postcss-cli": "^7.1.2",
```
amen-souissi, line 60: maybe ~ instead of ^ in this case

packages/css/.storybook/preview-head.html
```diff
@@ -6,7 +6,6 @@
   href="https://fonts.googleapis.com/css2?family=Roboto+Condensed:ital,wght@0,400;0,700;1,700&display=swap"
   rel="stylesheet"
 />
-<link href="./index.css" rel="stylesheet" />
```
amen-souissi, line 9: 👍

packages/css/package.json
```diff
@@ -5,12 +5,9 @@
   "author": "Decathlon <designsystem@decathlon.net>",
   "homepage": "https://github.com/Decathlon/vitamin-web#readme",
   "scripts": {
-    "dev": "npm run watch:storybook",
+    "dev": "start-storybook -p 6006 -s lib",
```
amen-souissi, line 8: Indeed it is better, while waiting for the problem with PostCSS 8. But we don't need lib anymore

packages/css/src/stories/components/buttons/buttons.stories.js
```diff
@@ -6,6 +6,8 @@ import secondaryReversed from './examples/secondary-reversed.html';
 import ghost from './examples/ghost.html';
 import conversion from './examples/conversion.html';
 
+import '~/src/index.css';
```
amen-souissi, line 9: I think this should be put in the storybook preview.js file ... ? it's more maintainable...

lerna.json
```diff
@@ -1,4 +1,6 @@
 {
   "packages": ["packages/*"],
-  "version": "0.0.1"
+  "version": "0.0.1",
+  "npmClient": "yarn",
```
amen-souissi, line 4: yarn instead of npm ? I agree, but is it validated by the rest of the team ?

packages/css/.storybook/main.js
```diff
@@ -9,4 +11,29 @@ module.exports = {
     '@storybook/addon-essentials',
     '@storybook/addon-storysource',
   ],
+  webpackFinal: async (config, { configType }) => {
```
amen-souissi, line 14: configType not used

amen-souissi: I think we no longer need to adapt the webpack configuration with PostCSS 7.  
Normally it should work without

amen-souissi (changes requested):
GG, minor changes

packages/css/package.json
jleveugle: yeah right ;)

lerna.json
jleveugle: It's a proposal :)

packages/css/.storybook/main.js
jleveugle: No, by default the loader for .css files is using style-loader and css-loader

Cf: https://github.com/storybookjs/storybook/blob/next/lib/core/src/server/manager/manager-webpack.config.ts#L116-L126

It's why I remove the default rule to use a new one, with postcss ;)

packages/css/src/stories/components/buttons/buttons.stories.js
jleveugle: Do you have an example ?

packages/css/.storybook/main.js
amen-souissi: this is the default config: console.log(config.module.rules[8].use);

```
[
  '/Users/AMEN/workspace/vitamin-web/node_modules/style-loader/dist/cjs.js',
  {
    loader: '/Users/AMEN/workspace/vitamin-web/node_modules/css-loader/dist/cjs.js',
    options: { importLoaders: 1 }
  },
  {
    loader: '/Users/AMEN/workspace/vitamin-web/node_modules/@storybook/core/node_modules/postcss-loader/src/index.js',
    options: { config: [Object] }
  }
]
```
remove node-modules, install and run storybook without this adaptation

packages/css/src/stories/components/buttons/buttons.stories.js
amen-souissi:
```
import '../src/index.css';

export const parameters = {
  actions: { argTypesRegex: '^on[A-Z].*' },
};

```

packages/css/.storybook/main.js
amen-souissi: I tested it and it works fine without :)

amen-souissi: for https://github.com/storybookjs/storybook/blob/next/lib/core/src/server/manager/manager-webpack.config.ts#L116-L126

I think there is a plugin system with the storybook webpack config

lerna.json
lauthieb: For me it's ok too because NPm is not ready for resolutions.

package.json
```diff
@@ -1,16 +1,32 @@
...
+    "url": "https://github.com/Decathlon/vitamin-web/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Decathlon/vitamin-web.git"
+  },
+  "license": "Apache-2.0",
+  "workspaces": [
+    "packages/*"
+  ],
```
lauthieb, line 19: Because you propose to use Yarn, we should add `engines` property inside `package.json` to force the user to have Yarn installed (https://classic.yarnpkg.com/en/docs/package-json/#toc-engines)
Also, we need to add this dependency and how to install's link inside our `README.md`

lauthieb (commented):
Thanks @jleveugle for this PR!!
Just a few suggestions!

lauthieb (changes requested):
Thanks @jleveugle! 
Just a few suggestions :)

packages/css/.storybook/main.js
jleveugle: Yeah, you are true.

My bad, I forgot to remove the first line, which remove the default css rule. It's working without, and the PR is up-to-date ;)

amen-souissi (approved):

lerna.json
thollander: Agreed that's a pretty well working configuration managing lerna with yarn workspaces.
(npm 7 included some workspace notion but I've never played with it )

thollander (approved):
Nice

package.json
```diff
@@ -23,52 +39,33 @@
...
-  "bugs": {
-    "url": "https://github.com/Decathlon/vitamin-web/issues"
-  },
-  "keywords": [
-    "design-system",
-    "ui-kit",
-    "libraries"
-  ],
-  "license": "Apache-2.0",
   "licenses": [
```
lauthieb, line 65: You can also move `licenses` to top as you did with `license` and other properties.

lauthieb (changes requested):

packages/css/.storybook/main.js
```diff
@@ -1,3 +1,5 @@
+const path = require('path');
```
amen-souissi, line 1: to remove

packages/css/.storybook/main.js
```diff
@@ -8,5 +10,5 @@ module.exports = {
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     '@storybook/addon-storysource',
-  ],
+  ]
```
amen-souissi, line 13: revert changes

lauthieb (approved):

PR #68, opened 2021-01-22
feat(@vtmn/css): add text input

gabbloquet:
## Pull Request checklist

## Description

Add text input CSS.

- No

## Other information

❤️ Thank you!


MatthiasCartelDkt (approved):
Good for me and open for contribute to this component !

lauthieb:
Ready for review 👍 
@Decathlon/front-end thanks in advance!

packages/sources/css/src/components/text-input/src/index.css
```diff
@@ -0,0 +1,69 @@
...
+      vtmn-rounded;
+  }
+
+  .vtmn-text-input:not([disabled]):hover {
+    box-shadow: inset 0 0 0 1px var(--vtmn-color_brand-digital);
+  }
+
+  .vtmn-text-input:not([disabled]):focus,
+  .vtmn-text-input:not([disabled]):active {
+    outline: none;
```
thollander, line 34: Removing outline is not a good accessibility practice... 

It should be kept and designed properly by designers. IMHO this element is quite tricky to tweak. I think designers will have to adapt to the difficulty of implementation

thollander (commented):
One small remark on a11y
Also, I would have put input-text instead of text-input, it more commonly used when you look at other design systems

packages/sources/css/src/components/text-input/src/index.css
lauthieb: But we have our proper outline.Like we did with the button.
How can you style outline with border radius?
I only find the trick of box-shadow inset...

lauthieb:
Thanks @thollander, can you send me examples for "input text"
The name comes from our website and we've seen a lot of examples:
- https://design-system.service.gov.uk/components/text-input/
- https://www.carbondesignsystem.com/components/text-input/code/
- https://design-system.pluralsight.com/components/textinput
- https://designsystem.gov.au/components/text-inputs/

But also, for the same component, a lot of text field:

- https://material.io/components/text-fields
- https://design.cms.gov/components/text-field/

thollander:
> Thanks @thollander, can you send me examples for "input text" please?
> The name comes from our website and we've seen a lot of examples:
> - https://design-system.service.gov.uk/components/text-input/
> - https://www.carbondesignsystem.com/components/text-input/code/
> - https://design-system.pluralsight.com/components/textinput
> - https://designsystem.gov.au/components/text-inputs/
> 
> But also, for the same component, a lot of text field:
> 
> - https://material.io/components/text-fields
> - https://design.cms.gov/components/text-field/
> 

Yep ok you've convinced me 😊
Let's keep it like this. 

I had this one in mind : https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%A7%AC-atoms-cc-input-text--default-story

Among others I don't remember but that's fine.

lauthieb:
Oh yes! I like a lot this UI kit by CleverCloud.
Okay, so let's keep text input but thanks again for your suggestion which could have been taken into consideration until we changed our website too!

packages/showcases/css/stories/components/text-input/examples/default.html
```diff
@@ -0,0 +1,10 @@
+<div class="block">
```
gabbloquet, line 1: I'm not a huge fan of "block". According to me it's generic and maybe that's the objective ?
I would prefer something like "text-input-wrapper" or "text-input-box", what do you think about that ?
(If you change take care to change it in the other files ;) )

packages/sources/css/src/components/text-input/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@vtmn/css-text-input",
+  "version": "0.0.0",
```
gabbloquet, line 3:
```suggestion
  "version": "0.0.1",
```
?

packages/sources/css/src/components/text-input/src/index.css
gabbloquet: Maybe we can report it to UX Team ? 
I am agree with @thollander, HTML / CSS Standards > ALL ;)

gabbloquet (commented):
Few little comments, thanks a lot 🚀

packages/sources/css/src/components/text-input/package.json
lauthieb: No because lerna will do it for us automatically thanks to our conventional commits. But if I made a commit with "feat", he will choose "0.1.0"

packages/showcases/css/stories/components/text-input/examples/default.html
lauthieb: This class is global for storybook and add a padding between each block. But yes I will use a tailwind css utility classes instead in every file I think.

packages/sources/css/src/components/text-input/src/index.css
lauthieb: UX/UI already take care of this and it passes WCAG level 2 accessibility criteria.
So I don't know why this 2px border is not accessible?

packages/sources/css/src/components/text-input/package.json
lauthieb: https://github.com/lerna/lerna/blob/main/commands/version/README.md#--conventional-commits

aarrtthhuurr:
What do you think of adding a minimum height for text-areas to avoid this behavior?
image

lauthieb (approved):

lauthieb:
@aarrtthhuurr Oh yes, you're right! Thanks for this suggestion.

packages/sources/css/src/components/text-input/src/index.css
gabbloquet: I was talking about css standard not about accessibility on that point, look good to me thanks :)

packages/showcases/css/stories/components/text-input/examples/default.html
gabbloquet: Oh that's true sorry, look good to me too ;)

PR #78, opened 2021-01-30
chore: fix stylelint bug

### LINKED ISSUE #48: bug: error on pre-commit hook from Husky with lerna + lint-staged
# Bug Report

We are facing a problem when we update a CSS file because the script `stylelint:fix` is failing when launched with lint-staged on pre-commit hook configured with Husky.

We should commit without add `--no-verify`

For example, to reproduce: 
- Go to the `main` branch and modify https://github.com/Decathlon/vitamin-web/blob/main/packages/sources/css/src/design-tokens/src/typography.css.
- Save your file
- `git add .`
- git commit -m "whatever" => KO:
image
- But it's OK if you only launch `yarn stylelint:fix`:
image

Thanks in advance for your help & thanks @gabbloquet for reporting me this too!

thollander:
- stylelint doesn't need to be launched as a lerna command
- rules should be globalized at root level
- bug was coming from the script to be launched in a package where command doesn't exist

Closes #48


lauthieb:
Oh, thanks for this @thollander!
Thank you also for these explanations ;-)

lauthieb (approved):

PR #230, opened 2021-03-21
feat(text-input): initialize Svelte

thollander:
## Description

Initialize the svelte version for text-input.
Needed to add Roboto-Condensed on CSS package of text-input. 

See #5

### DIFF
--- a/packages/sources/css/src/components/text-input/src/index.css
+++ b/packages/sources/css/src/components/text-input/src/index.css
@@ -6,7 +6,7 @@
 .vtmn-text-input_label,
 .vtmn-text-input,
 .vtmn-text-input_helper-text {
-  font-family: 'Roboto Condensed', 'system-ui', '-apple-system', sans-serif;
+  font-family: var(--vtmn-typo_font-family);
 }
 
 .vtmn-text-input,
--- a/packages/sources/css/src/components/text-input/src/index.css
+++ b/packages/sources/css/src/components/text-input/src/index.css
@@ -3,6 +3,12 @@
 @import '@vtmn/css-design-tokens/src/spacings';
 @import '@vtmn/css-design-tokens/src/typography';
 
+.vtmn-text-input_label,
+.vtmn-text-input,
+.vtmn-text-input_helper-text {
+  font-family: 'Roboto Condensed', 'system-ui', '-apple-system', sans-serif;
+}
+
 .vtmn-text-input,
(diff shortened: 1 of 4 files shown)


packages/sources/css/src/components/text-input/src/index.css
```diff
@@ -3,6 +3,12 @@
 @import '@vtmn/css-design-tokens/src/spacings';
 @import '@vtmn/css-design-tokens/src/typography';
 
+.vtmn-text-input_label,
+.vtmn-text-input,
+.vtmn-text-input_helper-text {
+  font-family: 'Roboto Condensed', 'system-ui', '-apple-system', sans-serif;
```
lauthieb, line 9:
```suggestion
  font-family: var(--vtmn-typo_font-family);
```

lauthieb: Added the CSS Custom Properties provided by our design tokens package to add Roboto.
For your information, Roboto Condensed is not the font that we have chosen for the text input component (due to lisibility).
Another information, following our survey #4, the font-family of the button will also change in the future from Roboto Condensed to Roboto.

packages/sources/svelte/src/components/VtmnTextInput.svelte
```diff
@@ -0,0 +1,77 @@
...
+   * Whether success should be shown or not
+   * @type {boolean}
+   */
+  export let showValid = true;
+
+  /**
+   * Whether error should be shown or not
+   * @type {boolean}
+   */
+  export let showError = true;
```
lauthieb, line 44: Why do you propose props for "show"?
As you can see for example with material-ui: https://material-ui.com/components/text-fields/
They only propose one prop "error", if the consumer tells that the error is true, so they directly show the error style.
I don't really understand why there is two props :)

lauthieb (changes requested):
Thank you Terence!
Just few suggestion & question.

packages/sources/svelte/src/components/VtmnTextInput.svelte
thollander: The use case would be : I want to have an error management but without having the "green/red" information for the user. 
But that's fine, I removed it, let's keep it simple for now :)

lauthieb (approved):

lauthieb:
Thanks a lot @thollander :tada:

PR #260, opened 2021-04-08
feat(@vtmn/css): add toggle component

GaspardMathon:
## Pull Request checklist

- [X] Make sure you are making a pull request against the **main branch** (left side).
- [X] Check your code additions will fail neither code linting checks nor unit test.

## Description

- Add toggle component in CSS
- Two toggle variants added -> an active and a disabled one.
- Add small and medium toggle
- Add icon on checked toggle

- No

## Other information

- Difficulty to change the label depending on the input state (need to set label text opacity to 0.5 when input is disabled)

❤️ Thank you!


crotru:
Hi !
I'm not really fan of the input inside its label. 
This way we'd prevent a proper text label that would target that swicth.
Hold tight. I'm looking for something a bit more a11y friendly...

packages/showcases/css/stories/components/toggle/examples/default.html
```diff
@@ -0,0 +1,7 @@
+<div class="block">
+  <label class="vtmn-toggle_switch">
+    <input type="checkbox" class="vtmn-toggle" />
+    <span class="vtmn-toggle_slider"></span>
+  </label>
+</div>
```
crotru, line 6: Maybe we should avoid nesting the input inside a dedicated label.
This will be problematic when implementing this next to  proper label like :
- [ ] Toggle me

"Toggle me" will have to be a label targeting '.vtmn-toggle'

Though it seems valid to have several label for the same input, it messes assistive technologies. 

Maybe we should dig something [this way](https://codepen.io/__sch__/pen/rNjYWxM?editors=1100) (keeping the input below the slider and keeping it clickable with a combination of opacity and / or pointer-events properties)

packages/sources/css/src/components/toggle/src/index.css
```diff
@@ -0,0 +1,74 @@
...
+  background-color: var(--vtmn-color_white);
+  transition: 0.4s;
+  border-radius: 50%;
+}
+
+.vtmn-toggle:not([disabled]) + .vtmn-toggle_slider:hover::before {
+  box-shadow: 0 0 0 rem(12px) rgba(0, 16, 24, 0.05);
+}
+
+.vtmn-toggle:not([disabled]) + .vtmn-toggle_slider:active::before {
```
crotru, line 56: `.vtmn-toggle` will receive `:active` state (not `.vtmn-toggle_slider`)

packages/sources/css/src/components/toggle/src/index.css
```diff
@@ -0,0 +1,74 @@
...
+  box-shadow: 0 0 0 rem(12px) rgba(0, 125, 188, 0.15);
+}
+
+.vtmn-toggle:checked + .vtmn-toggle_slider {
+  background-color: var(--vtmn-color_brand-digital);
+}
+
+.vtmn-toggle:checked + .vtmn-toggle_slider::before {
+  transform: translateX(rem(24px));
+}
```
crotru, line 74: We could have a `:focus` state though it might interfere a lot with the expected design.

We can at least add a `:focus-visible` pseudo-class (same declarations that for `:active` should be cool)

packages/sources/css/src/components/toggle/src/index.css
```diff
@@ -0,0 +1,74 @@
...
+/* The default slider */
+.vtmn-toggle_slider {
+  position: absolute;
+  cursor: pointer;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: var(--vtmn-color_grey-light-1);
+  transition: 0.4s;
```
crotru, line 30: We can transition only what's necessary:
`transition: background-color 0.4s;` here ...

packages/sources/css/src/components/toggle/src/index.css
```diff
@@ -0,0 +1,74 @@
...
+.vtmn-toggle_slider::before,
+.vtmn-toggle[disabled] + .vtmn-toggle_slider::before {
+  position: absolute;
+  content: '';
+  height: rem(24px);
+  width: rem(24px);
+  left: rem(4px);
+  bottom: rem(4px);
+  background-color: var(--vtmn-color_white);
+  transition: 0.4s;
```
crotru, line 48: ...and 
`transition: box-shadow 0.4s, transform 0.4s;` here 🙂

crotru (commented):
The switch looks cool. 
It could benefit a few tweaks tho 🙂

packages/showcases/css/stories/components/toggle/examples/default.html
GaspardMathon: Thanks for telling about accessibility, I didn't pay attention about this one.

packages/sources/css/src/components/toggle/src/index.css
GaspardMathon: Ok thanks :)

packages/sources/css/src/components/toggle/src/index.css
GaspardMathon: Ok thanks :) I put a focus-visible but it may change afterwards according to the designer choice

packages/sources/css/src/components/toggle/src/index.css
GaspardMathon: Ok thanks :)

packages/sources/css/src/components/toggle/src/index.css
GaspardMathon: Ok thanks :)

lauthieb (approved):
Thanks a lot @GaspardMathon for this great work!
You can see in details that I have had some refactoring there: https://github.com/Decathlon/vitamin-web/pull/260/commits/8f4b394b6eb708f21ebc65ea7b7ec5453647b898

z28kmorl (approved):

PR #261, opened 2021-04-09
feat(@vtmn/react): add text input component

gabbloquet:
## Pull Request checklist

## Description

Text Input feature, react part. (issue #5)

- No

## Other information

This PR is ✅, i'm waiting for you advices.

❤️ Thank you!

### DIFF
--- a/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
+++ b/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
@@ -36,6 +36,12 @@ export default {
       defaultValue: 'Helper text goes here',
       control: { type: 'text' },
     },
+    multiline: {
+      type: { name: 'boolean', required: false },
+      description: 'If the component is a textarea or an input.',
+      defaultValue: false,
+      control: { type: 'boolean' },
+    },
     disabled: {
       type: { name: 'boolean', required: false },
       description: 'The disabled state of the input.',
--- a/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
+++ b/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
@@ -12,12 +12,6 @@ export default {
       defaultValue: 'vtmn-input',
       control: { type: 'text' },
     },
-    className: {
-      type: { name: 'string', required: true },
-      description: 'The id of the input.',
-      defaultValue: 'vtmn-input',
-      control: { type: 'text' },
-    },
     labelText: {
       type: { name: 'string', required: true },
       description: 'The label of the input.',
--- a/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
+++ b/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
@@ -15,6 +15,12 @@ export default {
       defaultValue: 'vtmn-input',
       control: { type: 'text' },
     },
+    className: {
+      type: { name: 'string', required: true },
+      description: 'The id of the input.',
+      defaultValue: 'vtmn-input',
+      control: { type: 'text' },
+    },
     label: {
       type: { name: 'string', required: true },
       description: 'The label of the input.',
@@ -32,6 +38,24 @@ export default {
       description: 'The helper text message of the input.',
       defaultValue: 'Helper text goes here',
       control: { type: 'text' },
+    },
+    disabled: {
+      type: { name: 'boolean', required: false },
+      description: 'The disabled state of the input.',
+      defaultValue: false,
+      control: { type: 'boolean' },
+    },
+    valid: {
+      type: { name: 'boolean', required: false },
+      description: 'The validity state of the input.',
+      defaultValue: false,
+      control: { type: 'boolean' },
+    },
+    error: {
+      type: { name: 'boolean', required: false },
+      description: 'The error state of the input.',
+      defaultValue: false,
+      control: { type: 'boolean' },
     },
   },
   parameters: {
--- a/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
+++ b/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
@@ -19,7 +19,7 @@ export default {
       control: { type: 'text' },
     },
     placeholder: {
-      type: { name: 'string', required: true },
+      type: { name: 'string', required: false },
       description: 'The placeholder of the input.',
       defaultValue: 'Placeholder Text',
       control: { type: 'text' },
--- a/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
+++ b/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
@@ -1,7 +1,4 @@
-import React from 'react';
-
 import { VtmnTextInput } from '@vtmn/react';
-
 import { Story, Meta } from '@storybook/react';
 import { VtmnTextInputProps } from '@vtmn/react/src/components/VtmnTextInput';
 
--- a/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
+++ b/packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+
+import { VtmnTextInput } from '@vtmn/react';
+
+import { Story, Meta } from '@storybook/react';
+import { VtmnTextInputProps } from '@vtmn/react/src/components/VtmnTextInput';
+
+export default {
+  title: 'Components/VtmnTextInput',
+  component: VtmnTextInput,
+  argTypes: {
+    id: {
+      type: { name: 'string', required: true },
+      description: 'The id of the input.',
+      defaultValue: 'vtmn-input',
+      control: { type: 'text' },
+    },
+    label: {
+      type: { name: 'string', required: true },
+      description: 'The label of the input.',
+      defaultValue: 'Label',
+      control: { type: 'text' },
+    },
+    placeholder: {
+      type: { name: 'string', required: true },
+      description: 'The placeholder of the input.',
+      defaultValue: 'Placeholder Text',
+      control: { type: 'text' },
+    },
+    helperText: {
+      type: { name: 'string', required: false },
+      description: 'The helper text message of the input.',
+      defaultValue: 'Helper text goes here',
+      control: { type: 'text' },
+    },
+  },
+  parameters: {
+    backgrounds: { default: 'grey' },
+    actions: {
+      handles: ['mouseenter', 'click', 'focusin', 'focusout'],
+    },
+    design: {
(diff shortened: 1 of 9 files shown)


packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+  valid = false,
+  ...props
+}: VtmnTextInputProps) => {
+  return (
+    <>
+      <label className="vtmn-text-input_label" htmlFor={id}>
+        {label}
+        <input
+          className={`vtmn-text-input ${valid && 'vtmn-text-input--valid'}
+          ${error && 'vtmn-text-input--error'} ${className ? className : ''}`}
```
samedro, line 49: On pourrait utiliser la lib clsx, qui s'utiliserait de la manière suivante : 
```typescript
const classNames = clsx('vtmn-text-input', className, { 'vtmn-text-input--valid': valid }, { 'vtmn-text-input--error': error });

return (
   <input className={classNames} />
);
```
Histoire d'éviter toutes ces conditions

gabbloquet: Indeed thanks @samedro for you feedback :)

MatthiasCartelDkt (approved):

lauthieb:
Thanks @gabbloquet for this great contribution! I will review it quickly!

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+import clsx from 'clsx';
+
+import '@vtmn/css-text-input';
+
+export interface VtmnTextInputProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {
```
lauthieb, line 7: Could you please follow the same method that we use with `VtmnButtonProps` proposed by @FredericEspiau, please?

```suggestion
export interface VtmnTextInputProps
  extends React.ComponentPropsWithoutRef<'input'> {
```

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Label text linked to the input
+   * @type {string}
+   */
+  label: string;
```
lauthieb, line 12: It could be nice to follow the same props names that we already used in `VtmnTextInput.svelte` proposed by @thollander.
So for this example it's `labelText` instead of `label`

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Helper text to help the user
+   * @type {string}
```
lauthieb, line 16: It could be interesting to add `@defaultValue` in TSDoc for optional props and replace `@default` in TSDoc of `VtmnButton` by `@defaultValue` on the same occasion no?

See: https://tsdoc.org/pages/tags/defaultvalue/

```ts
   * @defaultValue false
```

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+  label,
+  placeholder,
+  valid = false,
+  ...props
+}: VtmnTextInputProps) => {
+  return (
+    <>
+      <label className="vtmn-text-input_label" htmlFor={id}>
+        {label}
+        <input
```
lauthieb, line 48: Why did you put the `<input>` inside the `<label>`?
See https://github.com/Decathlon/vitamin-web/blob/main/packages/showcases/css/stories/components/text-input/examples/default.html

lauthieb (changes requested):
Thanks @gabbloquet, few changes! :-)

packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+      description: 'The id of the input.',
+      defaultValue: 'vtmn-input',
+      control: { type: 'text' },
+    },
+    className: {
+      type: { name: 'string', required: true },
+      description: 'The id of the input.',
+      defaultValue: 'vtmn-input',
+      control: { type: 'text' },
+    },
```
lauthieb, line 20:
```suggestion
```

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+   * Whether input is successful or not
+   * @type {boolean}
+   */
+  valid?: boolean;
+
+  /**
+   * Whether input is in error or not
+   * @type {boolean}
+   */
+  error?: boolean;
```
lauthieb, line 30: Can you add a `placeholder` prop and make it optional, please?

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
```
lauthieb, line 31: Because we have chosen to have an `id` required, it could be nice to see it there as a prop types with TSDoc that explains it (because we use the "for" attribute of the required `<label>`)

packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+    label: {
+      type: { name: 'string', required: true },
+      description: 'The label of the input.',
+      defaultValue: 'Label',
+      control: { type: 'text' },
+    },
+    placeholder: {
+      type: { name: 'string', required: true },
```
lauthieb, line 28:
```suggestion
      type: { name: 'string', required: false },
```

lauthieb (changes requested):

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
gabbloquet: On the ID prop ? Not here ?

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
gabbloquet: Because i'm used to do that. And i think that's a HTML compliance. (Both are linked with id)
I can move it ;) 
https://developer.mozilla.org/fr/docs/Web/HTML/Element/Label

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
gabbloquet: On the helper ? I think that's not the case on svelte component 🤔

lauthieb: On each prop, yes but we can improve the doc like this no?

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
lauthieb: ? Just declare the `id` prop inside VtmnTextInputProps.

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
lauthieb: Nope, it's not the best practice.
The "for" already do the job to avoid having input inside label.

See this from W3C:

 image

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
gabbloquet: Indeed but actually i don't really know how to put as default 🤔

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+import clsx from 'clsx';
+
+import '@vtmn/css-text-input';
+
+export interface VtmnTextInputProps
+  extends React.ComponentPropsWithoutRef<'input'> {
```
lauthieb, line 7: So we also missed the `multilineˋ prop that we had inside our internal library to show a t'exagéra conditionnaly?

lauthieb (changes requested):

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
gabbloquet: I'm on it 🚀

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
gabbloquet: I made a suggestion ;) 
I look forward to your feedback :)

gabbloquet: I made a suggestion ;)
I look forward to your feedback :)

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+   * Label text linked to the input
+   * @type {string}
+   */
+  labelText: string;
+
+  /**
+   * Helper text to help the user
+   * @type {string}
+   * @default
+   *  Undefined and therefore not displayed by default
```
lauthieb, line 24:
```suggestion
   * @defaultValue undefined and therefore not displayed by default
```

lauthieb (changes requested):

packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+      description: 'The id of the input.',
+      defaultValue: 'vtmn-input',
+      control: { type: 'text' },
+    },
+    className: {
+      type: { name: 'string', required: true },
+      description: 'The id of the input.',
+      defaultValue: 'vtmn-input',
+      control: { type: 'text' },
+    },
```
lauthieb, line 20: As we did with Button and Icon, I think we don't need to display it in Storybook.
```suggestion
```

lauthieb (changes requested):

gabbloquet:
Thanks for you feedbacks 🙏🏻

lauthieb (approved):

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
FredericEspiau: Another important thing to note

```html
<label for="my-input" ...
<input id="my-input"...
```

Is supported by a larger number of screen readers

PR #270, opened 2021-04-15
docs(@vtmn/css): edit fonts import documentation

wlalele:
## Description

Edit the html import, replace the two html links by one that does the same thing.
Add another way of importing by using the css import rule (see [doc](https://developer.mozilla.org/fr/docs/Web/CSS/@import)).

- No


lauthieb (changes requested):
Thanks a lot @wlalele for this contribution!

Could it be possible for you to add these modifications inside our 9 `README` files that mentions this "Install fonts" paragraph inside this monorepo?

Thanks :-)

Issue #269, opened 2021-04-14
feat: add link component


GaspardMathon:
Feature request

Add link component

Documentation: https://www.decathlon.design/726f8c765/p/086ae8-link-beta/b/5496b9
Figma: https://www.figma.com/file/zDZIyayUlr1yTWrsi7cFoo/Vtmn---Web?node-id=1207%3A8898

Checklist:

- [x]  CSS (thanks @GaspardMathon)
- [x]  Web Components (thanks @GaspardMathon)
- [x]  React (thanks @gabbloquet)
- [x]  Svelte (thanks @thollander)
- [ ]  Vue

PR #276, opened 2021-04-17
feat(@vtmn/react): add toggle component

gabbloquet:
## Description

This is a pull request for Toggle implementation, for React. <VtmnToggle />
Linked issue: https://github.com/Decathlon/vitamin-web/issues/226 

## Does this introduce a breaking change?

- No

## Other information

I put only few props. I followed labelText naming as TextInput, and create VtmnToggleSize type as done with Button.
(Maybe it could be interesting to delete isDisabled props from VtmnTextInputProps interface ? :) )

❤️ Thank you!

PR #277, opened 2021-04-17
feat(@vtmn/react): add link component

gabbloquet:
## Description

This PR is for React implementation of Link component.
Issue related : #269 

- No

## Other information

I'm not really sure about props, href and target default values for storybook, if you have better ideas.
Firstly i wanted to put https://www.decathlon.fr 😃 

❤️ Thank you!


packages/showcases/react/stories/components/VtmnLink/VtmnLink.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+      },
+    },
+    isStandalone: {
+      type: { name: 'boolean', required: false },
+      description: 'If the component is a standalone or not.',
+      defaultValue: false,
+      control: { type: 'boolean' },
+    },
+    children: {
+      type: { name: 'string', required: false },
```
GaspardMathon, line 37: The 'children' represents the text of the link isn't it ? 
If it is, I think it should be required : 'true' because I don't see a link without text.

packages/sources/react/src/components/VtmnLink/types.ts
```diff
@@ -0,0 +1 @@
+export type VtmnLinkSize = 'small' | 'medium' | 'large';
```
GaspardMathon, line 1: I would be tented to make an enumeration here. To avoid setting size to 'small' but using VtmnLinkSize.SMALL instead.

packages/sources/react/src/components/VtmnLink/VtmnLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+   * Whether link is standalone or not.
+   * @type {boolean}
+   * @default false
+   */
+  isStandalone?: boolean;
+}
+
+export const VtmnLink = ({
+  isStandalone = false,
+  size = 'medium',
```
GaspardMathon, line 24: I would be tented to make an enumeration here. To avoid setting size to 'small' but using VtmnLinkSize.SMALL instead.

GaspardMathon (changes requested):

packages/sources/react/src/components/VtmnLink/types.ts
GaspardMathon: I made a request change (by accident) but it is more a suggestion ;)

packages/showcases/react/stories/components/VtmnLink/VtmnLink.stories.tsx
gabbloquet: It could be anything. Text, tag contains text, text and icon... ;) 
Anything in our component : 
<VtmnLink>
  </img>
</VtmnLink>

packages/sources/react/src/components/VtmnLink/types.ts
gabbloquet: Same comment  as in #276

packages/sources/react/src/components/VtmnLink/VtmnLink.tsx
gabbloquet: Same comment i think for consumption that's more difficult, they have to know the enumeration 🤔 
@lauthieb ? :D

packages/showcases/react/stories/components/VtmnLink/VtmnLink.stories.tsx
lauthieb: Yes so that's why I'm ok, children can be required. Because img, text, tag contains text, text and icon are children nodes

packages/sources/react/src/components/VtmnLink/types.ts
lauthieb: For propTypes like this I prefer union type (https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)
Because that's the point, we need to know what are the different types possible and I used to see that on React components.

packages/sources/react/src/components/VtmnLink/VtmnLink.tsx
lauthieb: Yes I understand the argument, even for library side only, it could be nice to don't have "magic words" like "medium" and avoid bugs by securing with enum. I will investigate.

packages/showcases/react/stories/components/VtmnLink/VtmnLink.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+      description: 'The href of the link.',
+      defaultValue: '#',
+      control: { type: 'text' },
+    },
+    target: {
+      type: { name: 'string', required: true },
+      description: 'The target of the link.',
+      defaultValue: '_blank',
+      control: { type: 'text' },
+    },
```
lauthieb, line 20: Target mustn't be required. Indeed, when we need to have an internal link, in HTML we don't specify it.
```html
<a href="/page-2.html">Go to next page</a>
```

lauthieb (approved):

GaspardMathon (approved):

packages/showcases/react/stories/components/VtmnLink/VtmnLink.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+  component: VtmnLink,
+  argTypes: {
+    href: {
+      type: { name: 'string', required: true },
+      description: 'The href of the link.',
+      defaultValue: '#',
+      control: { type: 'text' },
+    },
+    target: {
+      type: { name: 'string', required: true },
```
lauthieb, line 16:
```suggestion
      type: { name: 'string', required: false },
```

lauthieb (approved):

Issue #257, opened 2021-04-07
feat: add icons in text input component


lauthieb:
**Feature request**

Add icons in text input component in input and also, in error message.

Documentation: https://www.decathlon.design/726f8c765/p/31121d-text-input
Figma: https://www.figma.com/file/zDZIyayUlr1yTWrsi7cFoo/Vtmn-Web?node-id=150%3A4893

Checklist:
- [ ] CSS
- [ ] Web Components
- [ ] React
- [ ] Svelte
- [ ] Vue

PR #314, opened 2021-05-10
feat(text-input): add icon for react

gabbloquet:
## Description

This is pull request to add Icons on test input for React.
Linked issue : https://github.com/Decathlon/vitamin-web/issues/257 

- No

## Other information

❤️ Thank you!


lauthieb (changes requested):
Thanks @gabbloquet, a remark about default value for icon. Icon is not required and an input can be there without any icon inside. (so without any `VtmnIcon` component)

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -70,34 +79,39 @@ export const VtmnTextInput = ({
...
+                { 'vtmn-text-input--valid': valid && !disabled },
+                { 'vtmn-text-input--error': error && !disabled },
+              )}
+              id={id}
+              type="text"
+              placeholder={placeholder}
+              disabled={disabled}
+              {...props}
+            />
+            <VtmnIcon value={icon} />
```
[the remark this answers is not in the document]
gabbloquet: Indeed i will add it, actually i didn't put it because <VtmnIcon /> already contains it but indeed that's safer to add it 👌🏻

packages/showcases/react/stories/components/VtmnTextInput/VtmnTextInput.stories.tsx
```diff
@@ -30,6 +31,15 @@ export default {
       defaultValue: 'Helper text goes here',
       control: { type: 'text' },
     },
+    icon: {
```
lauthieb, line 34: What do you think about rename it to "iconValue" instead of "icon"?
It could clearer for user to add a string instead of a `VtmnIcon` Component inside this prop no?

lauthieb: @gabbloquet ?

gabbloquet: I think it's overkill because the only icon props its that one.
It could be interesting to have other advice on it, @RomainJoly , @gaspard-mathon, what's your advice on it ?
`icon` VS `iconValue` props ?

lauthieb: Yes I understand. But when I see "icon" for prop name, I'm waiting for passing a `<Icon>` element as prop instead of a value

RomainJoly: For me it is ok to have props name "icon".The prop name "icon" in the button component of Vitamin is waitting a string.

gaspard-mathon: I don't see any problem to call the prop "icon". The consumers can see what type of data is required in the showcase.

lauthieb: Okay thanks @RomainJoly !

lauthieb:
for me it's ok! Thanks @gabbloquet for this contribution!!!

gabbloquet:
with pleasure @lauthieb, sorry for this lack of responsiveness

PR #358, opened 2021-06-14
feat(@vtmn/vue): add link component

lauthieb:
## Pull Request checklist

## Description

Add link component for Vue.

- No

## Other information

❤️ Thank you!

### DIFF
--- a/packages/sources/vue/src/components/VtmnLink/VtmnLink.vue
+++ b/packages/sources/vue/src/components/VtmnLink/VtmnLink.vue
@@ -1,0 +1,41 @@
+<script lang="ts">
+import '@vtmn/css-link/dist/index.css';
+import { reactive, computed, defineComponent } from 'vue';
+
+export default /*#__PURE__*/ defineComponent({
+  name: 'VtmnLink',
+  props: {
+    size: {
+      type: String,
+      default: 'medium',
+      validator: (val: string) => ['small', 'medium', 'large'].includes(val),
+    },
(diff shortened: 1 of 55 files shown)


packages/sources/vue/src/components/VtmnLink/VtmnLink.vue
```diff
@@ -0,0 +1,41 @@
...
+  },
+  setup(props) {
+    props = reactive(props);
+
+    return {
+      classes: computed(() => ({
+        'vtmn-link': true,
+        [`vtmn-link_size--${props.size}`]: true,
+        'vtmn-link--standalone': props.isStandalone,
+        'vtmn-link--icon-along': props.hasIconAlong,
```
gaspard-mathon, line 30: I think the component needs more logic because if the link is inline it can't have an icon along.

gaspard-mathon (changes requested):
Add more logic for inline case

packages/sources/vue/src/components/VtmnLink/VtmnLink.vue
lauthieb: I take the same algorithm as Svelte, React and Web Components, so yes we will need to refactor this in all stack

lauthieb: @GaspardMathon we still need to add the same algorithm to other frameworks.

PR #516, opened 2021-08-14
chore: nx as monorepo manager

thollander:
## Pull Request checklist

## Description

As the project grows (multiple packages, multiple Storybook), it doesn't scale well. 
All tasks are run across all projects everytime whereas it's not needed sometimes. 
Adding NX to the project will benefit of : 
- computation caching (not rebuilding if it has already been cached)
- smart rebuilds of affected projects (great to build only what has been changed in a commit)

For now, I've not removed Lerna which will keep managing versions publishing. 

I've also refactored `yarn lint = yarn prettier + yarn stylelint`. 

We may also consider in the future running `affected` tasks on merging on `main`. 
I kept it simple for now to see how it's going on on PR only and then we'll see on `main` to compare with `HEAD^` commit. 

Nx comes with a free cloud that shows results of commands. It would be interesting to comment the PR with such link in a next future. 

We'll also need to add `NX_CLOUD_AUTH_TOKEN` as a secret in the repository in order to have a global cache mechanism on nx cloud and build reports exported in it. 

We can already see some benefits of caching between `install` phase vs `build` phase.
See on another PR : 
image
🤔 Actually, why is it building while installing dependencies ? (poke @lauthieb)
Benefits will mostly be seen when this PR will be merged as the `affected` option here doesn't do anything. 

- No

❤️ Thank you!


thollander:
`yarn dep-graph`

image

nx.json
```diff
@@ -0,0 +1,46 @@
+{
+  "npmScope": "vitamin-web",
```
lauthieb, line 2: Does npmScope should'nt be `@vtmn`?

lauthieb (commented):
Thanks @thollander!!
I'll test it tomorrow.

nx.json
thollander: Don't really know 😕 I'll switch to `vtmn` anyway you're right


[Docs](https://nx.dev/latest/react/core-concepts/configuration) are not really clear : 


```
NPM Scope

Tells Nx what prefix to use when generating library imports.
```

thollander: image

I guess it's the scope without the `@`

jleveugle:
We should replace lerna completly before merging this ! Great start !@thollander ;)

lauthieb:
Thanks a lot @thollander for this contribution! It's a great start to improve our build chain.
I didn't understand your question about build & install, can you please reformulate?

lauthieb:
@thollander oh yes you're right, because we already have a `"prepare"` script that runs `yarn build` so we don't need to have a build step inside our GitHub actions workflows, you can delete it! Thanks :)

lauthieb (approved):

thollander:
> @thollander oh yes you're right, because we already have a `"prepare"` script that runs `yarn build` so we don't need to have a build step inside our GitHub actions workflows, you can delete it! Thanks :)

I would be for removing it from the "prepare" script instead of removing the build in the pipeline... but let's make that in another PR. 
I guess we can go further with this one ? Have to fix conflicts quite often while we wait for it to be merged

GaspardMathon (approved):

lauthieb:
@thollander you're right :-), here is the PR: https://github.com/Decathlon/vitamin-web/pull/592

lauthieb:
Ping @rlopes27, feel free to join us to this discussion if you know monorepo management tools like Lerna & nx! :)

lauthieb (approved):

lauthieb:
Just refactored as we said this afternoon.
Thanks again @thollander & @jleveugle!
Now it's okay for me! Let's merge it 🎉

PR #651, opened 2021-10-04
fix(@vtmn/react): TS compilation error 

### LINKED ISSUE #650: bug(@vtmn/react): Typescript warnings in the console when `yarn build`
# Bug Report

Hello,
There are Typescript warnings in the console when `yarn build` on `@vtmn/react` package.

Capture d’écran 2021-10-04 à 12 37 14

**Steps to reproduce**

`yarn && cd packages/sources/react && yarn build`

Thanks!

blemoine:
## Pull Request checklist

- [X]  Check your code additions will fail neither code linting checks nor unit test.
- [X] Check PR name, it must follow the https://www.conventionalcommits.org pattern

## Description

Fix compilation error for React components ; 

* React.memo doesn't support components returning an array,
  hence rollbacking to fragment syntax.
* `input`s and `textarea`s don't have the same valid HTML attributes.
   As the component uses `multiline` to know when to render one or the other,
   I explicited in the type that if `multiline` is true,
   then the expected attributes are of textarea,
   and that if `multiline` is false, the expected attributes are of input.
* VtmnToggle overrides `size` with its own semantic, 
   it implies that VtmnToggleProps doesn't fully extends HTMLInputProps

Closes #650 

- Yes

User of VtmnTextInput  using `multiline=true` and explicitly writing a callback expecting a HTMLInputElement may have to change the type to HTMLTextareaElement.  While this is a "breaking change" for the user - as they have to change their code - it's really a fix, because they were really receiving a HTMLTextareaElement, it was jsut badly typed. 

## Other information

Just a piece of opinion: I suggest to transform TS warns to TS errors (preventing the merge). Of the 3 problems above, at least 2 were real problems that could lead to users bugs.  (I'm not 100% sure about `React.memo`, if the problem really lies with `memo` Or with the type declaration from facebook)

❤️ Thank you!

### DIFF
--- a/packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
+++ b/packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
@@ -79,7 +79,6 @@ export const VtmnTextInput = ({
   icon = undefined,
   identifier,
   labelText,
-
   placeholder,
   valid = false,
   ...props
--- a/packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
+++ b/packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
@@ -5,8 +5,7 @@ import '@vtmn/css-text-input/dist/index-with-vars.css';
 import { VtmnIcon } from '../VtmnIcon';
 import { VitamixId } from '@vtmn/icons/dist/vitamix/font/vitamix';
 
-export interface VtmnTextInputProps
-  extends React.ComponentPropsWithoutRef<'input'> {
+type VtmnTextInputAdditionalProps = {
   /**
    * ID of the input
    * @type {string}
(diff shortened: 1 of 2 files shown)


GaspardMathon (approved):

packages/sources/react/src/components/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -69,56 +79,58 @@ export const VtmnTextInput = ({
   icon = undefined,
   identifier,
   labelText,
-  multiline = false,
+
```
lauthieb, line 82:
```suggestion
```

lauthieb (approved):

lauthieb:
Thank you very much @blemoine for these fixes and their explanations. Really appreciated. ❤️

PR #677, opened 2021-10-13
fix(@vtmn/vue): typescript build init

### LINKED ISSUE #440: Issues with Vue 3 + Typescript project
Hi guys, I'm building a new front-end project with Vue 3 + Typescript and I encounter 2 problems (or misunderstanding).

1) I get Typescript errors when importing components from the design system. As indicated in the documentation I import my component like this: `import { VtmnButton } from '@ vtmn/vue/dist/esm/VtmnButton'`

This generates the following error:

```
Could not find a declaration file for module '@vtmn/vue/dist/esm/VtmnButton'. '/Users/THOMAS/Projects/finance-front-catalog/node_modules/@vtmn/vue/dist/esm/VtmnButton.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/vtmn__vue` if it exists or add a new declaration (.d.ts) file containing `declare module '@vtmn/vue/dist/esm/VtmnButton';`
```

I could force Typescript to ignore this error or to implicitly type it as any but it is not a lasting solution.

I also wonder why not be able to import the components by doing instead `import { VtmnButton } from '@vtmn/vue'`? It makes more sense to me than having to go to the`dist/esm` directory.

2) By using the import method mentioned above I am also obliged to install the CSS package associated with the component (Button). Here is the error:

```
 ERROR  Failed to compile with 1 error

This dependency was not found:

* @vtmn/css-button/dist/index.css in ./node_modules/@vtmn/vue/dist/esm/VtmnButton.js

To install it, you can run: npm install --save @vtmn/css-button/dist/index.css
```

I already load upstream the global CSS of the design system, I have the impression that it is duplicative to have to load in addition the CSS of each component and it might not be obligatory.

Maybe it is missing some declarations or types inside the components ? It also possible that I'm missing or misunderstanding something. In both case I am sure it is related to Typescript.

Thanks for your help

**Versions**

- @vtmn/vue: 0.6.2

lauthieb:
## Pull Request checklist

## Description

Following suggestions from @Wizyma, thanks a lot!
Close #440

- No

## Other information

❤️ Thank you!


Wizyma:
hey @lauthieb 
So after some investigation (and getting to know vue in the same time ahah :p) i reviewed a bit the config, and for the code itself i looked on how others `vue3` lib were built. 

if we look at [Equal](https://github.com/quatrochan/Equal) for example we can see that overall our conf looks the same, but the components how they are built, how the shim is writen and how the `index.ts` looks is quite difference from ours, so i tried going with this approach and it seems to work quite well, i still have some issues with the typings (it seems to be a real pain point :D) 

And as for storybook it was a cache issue and we need to import the css file from the vue lib in storybook : `import '@vtmn/vue/dist/style.css'`

https://user-images.githubusercontent.com/29448654/137215875-8b0c0eb0-289e-425e-a83a-63b9d9cd9cd9.mov

So to summarize components declarations now look like
```typescript
import { App as Application } from 'vue';
import YOUR_COMPONENT from './YOUR_COMPONENT.vue';

YOUR_COMPONENT.install = (Vue: Application) => {
  Vue.component(YOUR_COMPONENT.name, YOUR_COMPONENT);
};

export default YOUR_COMPONENT;
```

`shim-vue.d.ts` : 
```typescript
declare module '*.vue' {
  import { defineComponent } from 'vue'
  const Component: ReturnType<typeof defineComponent>
  export default Component
}
```

`index.ts`
```typescript
/* eslint-disable */
import { App as Application, Plugin } from 'vue';
import * as components from './components/index';

function install(Vue: Application) {
  // tslint:disable-next-line: forin
  for (const component in components) {
    // @ts-expect-error
    Vue.component(components[component].name, components[component])
  }
}

export default install;

export { default as VtmnButton } from './components/VtmnButton'
export { default as VtmnCheckbox } from './components/VtmnCheckbox'
export { default as VtmnLink } from './components/VtmnLink'
export { default as VtmnRadioButton } from './components/VtmnRadioButton'
export { default as VtmnTextInput } from './components/VtmnTextInput'
export { default as VtmnToggle } from './components/VtmnToggle' 
```

the vite config 
```typescript
import typescript from '@rollup/plugin-typescript'; // new import added

module.exports = defineConfig({
  plugins: [
    ...plugions
    typescript({ sourceMap: true })
  ],
  build: {
     ...
  },
});

```

I still need to spend more time on the types part but thats a good step forward on improving the tooling !!

Wizyma:
hey @lauthieb  ! 
yep it should ! And do not forget to launch storybook with `--no-manager-cache`  and to add the styles import to it !

Wizyma:
> Hmm... Is it possible to have the same behavior as we already have with `@vtmn/react`, the fact that we doesn't need to explicitly import CSS styles. They are scoped inside each components.

I will see what i can do to not break the current behavior !

lauthieb:
Hi @Wizyma, thanks a lot for your suggestions! I will keep your solution even if there are some cons.
I've just made your propositions and it works. Can you please review my work by seeing the new `README.md` also?
You can test easily by launching: `yarn start:vue` and try to modify a Vue story and some Vue sources.
Thanks!

Issue #526, opened 2021-08-23
refactor(@vtmn/css): change rem/px units for some properties


GaspardMathon:
**Subject**

In order to have a consistent Design System, we must define which properties are dependent of the root font size.
Typically, margin, padding are dependent of the root font size whereas borders, shadows are not.

lauthieb:
Completely agree

GaspardMathon:
What do you think about this stackoverflow answer ? 

```
Respecting User Preferences
Browser zoom has changed a lot over the years. Historically many browsers would only scale up font-size, but that changed pretty rapidly when websites realized that their beautiful pixel-perfect designs were breaking any time someone zoomed in or out. At this point, browsers scale the entire page, so font-based zooming is out of the picture.

Respecting a user's wishes is not out of the picture. Just because a browser is set to 16px by default, doesn't mean any user can't change their preferences to 24px or 32px to correct for low vision or poor visibility (e.x. screen glare). If you base your units off of rem, any user at a higher font-size will see a proportionally larger site. Borders will be bigger, padding will be bigger, margins will be bigger, everything will scale up fluidly.

If you base your media queries on rem, you can also make sure that the site your users see fits their screen. A user with font-size set to 32px on a 640px wide browser, will effectively be seeing your site as shown to a user at 16px on a 320px wide browser. There's absolutely no loss for RWD in using rem.
```

@lauthieb @xavhan 

Basically, we shouldn't put borders padding in px and the font with rem in order to keep the design consistent if the user chooses a different root size from his browser. 

Concrete example, If you set your default browser font size at 50px, the font size root will cascade for all the rem props but the borders, that can be 1px wide will remains 1px wide if we don't use rem on border property. So The user will probably not see the borders.

GaspardMathon:
I have tested the behaviour of all the components with 500% zoom in, I saw nothing anoying about sizes.
I would say to let the units in `px` while using the `rem()` postcss function.

@lauthieb any feedback on this ?

PR #730, opened 2021-11-05
feat(@vtmn/css): add chip component

GaspardMathon:
## Pull Request checklist

- [X] Check your code additions will fail neither code linting checks nor unit test.
- [X] Check PR name, it must follow the https://www.conventionalcommits.org pattern
- [X] If it's a new component in CSS, ask for design review

## Description

- Add the chip component with all variants
- Improvement on the badge component (fix font-weight, line-height and min-width)
- Fix card linear-gradient
- Fix hover color for the text-input box-shadow

- No

❤️ Thank you!


packages/sources/css/src/components/badge/src/index.css
```diff
@@ -5,8 +5,11 @@
 .vtmn-badge {
   font-family: var(--vtmn-typo_font-family);
   font-size: var(--vtmn-typo_text-3-font-size);
+  font-weight: 500;
```
lauthieb, line 8: Sorry it's impossible because we require only 400 and 700 in all our README. We need to convince designers that it's a bad practice to have a weight of 500 just for one little component as the badge.
I vote for 400 :)

packages/sources/css/src/components/text-input/src/index.css
```diff
@@ -38,7 +38,8 @@ textarea.vtmn-text-input {
 }
 
 .vtmn-text-input:not(:disabled):not(:focus):not(textarea):hover {
-  box-shadow: inset 0 0 0 rem(1px) var(--vtmn-semantic-color_border-active);
+  box-shadow: inset 0 0 0 1px var(--vtmn-semantic-color_border-inactive),
+    0 0 0 3px var(--vtmn-semantic-color_hover-primary);
```
lauthieb, line 42:
```suggestion
  box-shadow: inset 0 0 0 rem(1px) var(--vtmn-semantic-color_border-inactive),
    0 0 0 rem(3px) var(--vtmn-semantic-color_hover-primary);
```

Let's keep `rem()` function for the moment. This refactoring will be part of #526

lauthieb (changes requested):

lauthieb (changes requested):
Thanks @GaspardMathon for this work on chip component!
I have two remarks until it's ok:

1/ The focus state is not okay for unselected chip

Not focused:
Capture d’écran 2021-11-09 à 10 38 40

Focused
Capture d’écran 2021-11-09 à 10 38 47
It looses its border :/ 

2/ The behavior is not like we said together for the close icon. If a close icon is there, the div has no role button but the close icon is the only button.

Feel free to ping me if you need more information.

Thanks.

lauthieb:
Hi @GaspardMathon, I let you check my suggestions :)

packages/sources/css/src/components/badge/src/index.css
GaspardMathon: I went back for the token regular font-weight

packages/sources/css/src/components/text-input/src/index.css
GaspardMathon: I put back rem for every px size

lauthieb (approved):

lauthieb (approved):

Issue #719, opened 2021-10-29
refactor(@vtmn/react): VtmnTooltip position typings are not accessible from other TypeScript projects


AntoineDeloy:
Current Behaviour :
The VtmnTooltip contanins a VtmnTooltipPosition which is an enum, it's not ideal because we need to declare this enum in our project to be able to use it and send it to the VtmnTooltip component as a valid props.

Expected Behaviour : 
I suggest to replace the VtmnTooltipPosition attribute as a exported type, not an. enum, to be able to send recognized string value directly by the component, and not being forced to declare the enum in our architecture.
This behaviour is indeed the way that other vtmn react component works, their attributes are all in type (you can check VtmnIcon, Button, etc.)

lauthieb:
@AntoineDeloy, as we've seen together on Slack, I've now assigned you, feel free if you have question regarding your contribution/onboarding on vitamin-web :tada:

Wizyma:
Hey @AntoineDeloy !

the type `VtmnTooltipProps` is exposed from the library so you should be able to use the `tooltip` wich is available in the `VtmnTooltipProps` interface like:

```typescript
import { VtmnTooltipProps } from '@vtmn-web/react';

type tooltip = Pick<VtmnTooltipProps , 'tooltip'>
```

CF: [typescript utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys)

AntoineDeloy:
Hello ! Sorry for the delay, I had a lot a of work on Green Design team, I'm coming back to this issue and I don't see why it would be a breaking as it is really blocking right now for users because they need to declare the enum in their architecture, so changing  it won't fail the compilation as if they still send string value the component will still understand the VtmnPosition props

Wizyma:
Hey Hey @AntoineDeloy ! you are suppppeer right !

PR #748, opened 2021-11-18
refactor(@vtmn/react): VtmnTooltipPosition props from enum to type

AntoineDeloy:
## Pull Request checklist

## Description

Refacto of the position tooltip props, from an enum to a type to comply with current vtmn react component behaviour, as it will be easier to modify components props from sending the position directly as a string value from an authorized range declare in the position exported Type 

As seen in my issue https://github.com/Decathlon/vitamin-web/issues/719 i don't think this is a breaking change but maybe I'm wrong ?

- No

## Other information

The position of tooltip is now modifiable easily by users because before as it was an enum, developers must create the enum in their local architecture, now it complies with the behaviour if current vtmn react components, the props which must receive certain string values are declared as a Type.

❤️ Thank you!

### DIFF
--- a/packages/sources/react/src/components/VtmnTooltip/types.ts
+++ b/packages/sources/react/src/components/VtmnTooltip/types.ts
@@ -1,7 +1,7 @@
 export type VtmnTooltipPosition =
   | 'bottom-left'
   | 'bottom'
-  | 'bottom - right'
+  | 'bottom-right'
   | 'left'
   | 'right'
   | 'top-left'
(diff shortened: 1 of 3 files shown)


packages/sources/react/src/components/VtmnTooltip/types.ts
```diff
@@ -1,10 +1,9 @@
...
-  LEFT = 'left',
-  RIGHT = 'right',
-  TOP_LEFT = 'top-left',
-  TOP = 'top',
-  TOP_RIGHT = 'top-right',
-}
+export type VtmnTooltipPosition =
+  | 'bottom-left'
+  | 'bottom'
+  | 'bottom - right'
```
lauthieb, line 4:
```suggestion
  | 'bottom-right'
```

lauthieb (changes requested):

lauthieb:
Thanks Antoine, I will check the problem!

AntoineDeloy:
@lauthieb I don't understand why the build still fails, do you think in's related to the package.json or something similar ? Cause i did not change anything that could throw an error in showcases and on my local the build is not failing at all

PR #758, opened 2021-11-22
feat(@vtmn/react): add VtmnAlert component

AntoineDeloy:
## Pull Request checklist

## Description

Pull Request to add the React alert component

## Other information

I needed to change the CSS of the #root storybook element from display: flex to display: inline-block cause due to the max-width: 40% of the vtmn-alert class, this was truncating the component width and it was not displaying well in the storybook.

Is it a problem ? The behaviour of the storybook is still the same, but now the elements are on the top left of the root element. Of course I would be happy if you found another solution !

❤️ Thank you!


AntoineDeloy:
Hello @lauthieb and @GaspardMathon this implementation is working but I need your advide on it, because there is 2 ways of implementing it  : 
The hide/show intelligence is the parent component (so not in the design system), it's easier but the the timout and queue intelligence have to be implemented in each of the future applicaiotn that will use it, what do you want about that ? Leave our users decide their implementation or restrain them in a specific standard ?

The implementation of timeout / queue / hide & show can be implemented in the vtmnAlert component, right now it's not testable in the storybook becasue there is a problem with fallback function that can't be used in the child component of stories' template.

What do you think about what i've done right now ? And what is your decision about the behaviour of this component ?
Thanks in advance !

lauthieb:
I've just converted your PR in draft about what we said yesterday.
Feel free to tell us when it's ok. Thanks a lot!

PR #783, opened 2021-11-30
refactor(@vtmn/css): fix width fixed value

GaspardMathon:
## Pull Request checklist

## Description

- The cards used to have a fixed width value, they are now responsive.

- No

❤️ Thank you!


packages/sources/css/src/components/card/src/index.css
```diff
@@ -122,13 +119,13 @@
...
+  background: linear-gradient(
     180deg,
     rgb(0 16 24 / 0%) 20%,
     var(--vtmn-semantic-color_background-primary-reversed) 100%
   );
 }
 
 .vtmn-card_variant--full-image > .vtmn-card_content--opaque {
-  background-color: rgb(0 16 24 / 60%);
+  background: rgb(0 16 24 / 60%);
```
lauthieb, line 130: why not keeping `background-color` in this case?

packages/sources/css/src/components/card/src/index.css
```diff
@@ -122,13 +119,13 @@
   height: auto;
   border-radius: 8px;
   color: var(--vtmn-semantic-color_content-primary-reversed);
-  background-color: linear-gradient(
+  background: linear-gradient(
```
lauthieb, line 122: why not keeping `background-color` in this case?

packages/sources/css/src/components/card/src/index.css
```diff
@@ -105,6 +101,7 @@
 
 .vtmn-card_variant--full-image img {
   height: 100%;
+  min-width: rem(200px);
```
lauthieb, line 104: Why?

packages/showcases/css/stories/components/card/examples/full-image.html
```diff
@@ -16,7 +16,7 @@
   </div>
 </div>
 
-<div class="block">
+<div class="block" style="width: 50%">
   <div class="vtmn-card vtmn-card_variant--full-image">
     <div class="vtmn-card_image">
       <img
```
lauthieb, line 22: There's a problem with the second full image in dark mode:
Capture d’écran 2021-11-30 à 23 13 56

lauthieb (changes requested):

packages/sources/css/src/components/card/src/index.css
GaspardMathon: nope it doesn't work with the second background-color, kind of bug I think.

packages/sources/css/src/components/card/src/index.css
GaspardMathon: Because it bugs with the previous background-color, idk why

packages/sources/css/src/components/card/src/index.css
GaspardMathon: To avoid having the content going out of the image, but maybe I can refacto this

packages/sources/css/src/components/card/src/index.css
GaspardMathon: Ok, after research, it seems like we must set a linear-gradient with background, not background-color : https://developer.mozilla.org/fr/docs/Web/CSS/gradient/linear-gradient()

packages/sources/css/src/components/card/src/index.css
GaspardMathon: Indeed, we need to discuss about this. Tomorrow ? :)

lauthieb (approved):

PR #791, opened 2021-12-08
feat(@vtmn/react): add alert component

AntoineDeloy:
## Pull Request checklist

## Description

Pull Request to add the React alert component

## Other information

Tried to make the component a easy to use as possible, the intelligence of showing / hiding the alert needs to be in the application that uses this component and defined by the dev, the storybook shows the standard behaviour of this component

❤️ Thank you!


packages/sources/react/src/components/VtmnAlert/VtmnAlert.tsx
```diff
@@ -0,0 +1,73 @@
...
+          `vtmn-alert_variant--${type}`,
+          { show: showAlert },
+          className,
+        )}
+      >
+        <article className="vtmn-alert_content">
+          <div className="vtmn-alert_content-title">
+            {title}
+            {onClose && (
+              <span className="vtmx-close-line" onClick={() => onClose} />
```
bastienterrier, line 57: Don't we want to call the `onClose` function?
I would have write `onClick={() => onClose}` or `onClick={onClose}` as we don't need `this` context

packages/sources/react/src/components/VtmnAlert/VtmnAlert.tsx
```diff
@@ -0,0 +1,73 @@
+import * as React from 'react';
+import { VtmnAlertType } from './types';
+import clsx from 'clsx';
+
+export interface VtmnAlertProps
+  extends React.ComponentPropsWithoutRef<'dialog'> {
+  /**
+   * The alert's title
+   */
+  title: string;
```
bastienterrier, line 10: Is the `title` mandatory? I have a use-case when I don't want to show title, only some content. 
Is it necessary for accessibility purpose?

packages/sources/react/src/components/VtmnAlert/VtmnAlert.tsx
AntoineDeloy: You're right !

AntoineDeloy:
For me, the close icon should not be optional, and be always visible as the VtmnAlet component should always be closable.
Why ? 
An Alert should not be intrusive, the user should not be forced to wait the end of the alert in any way, and should be able to close it as soon as he clicks anywhere ON the component, the close icon would always be here just for the information of "okay, this component is closable if I want it to"

lauthieb:
Thanks a lot @AntoineDeloy for this PR!
We will try to see why the `onClose` method is not called later on Storybook.
Have a nice evening.

AntoineDeloy:
Still some PR to come don't worry !

Issue #717, opened 2021-10-26
feat: add icon component


lauthieb:
**Feature request**

Add icon component.

Documentation: https://www.decathlon.design/726f8c765/p/91dc94-iconography

Checklist:
- [x] React (thanks @gabbloquet)
- [x] Svelte (thanks @Tlahey)
- [x] Vue (thanks @ThomasLaigneauDecathlon)

ThomasLaigneauDecathlon:
Starting to work on the Vue implementation

PR #850, opened 2022-01-04
feat(@vtmn/react): update icon variants

crotru:
## Pull Request checklist

## Description

This is an update to react [icon component](https://github.com/Decathlon/vitamin-web/issues/717) variants
- **It allows icons to have an inherited color.**
_Why_ : This makes sense as icons doesn't have color restrictions, and when associated with text, an icon should be able to hold the same color.
_How_ : This PR propose a `neutral` variant keyword, when stated outputted inline color value is `inherit`

- **It changes the variant keyword `danger` as `negative`**
and output color value as `var(--vtmn-semantic-color_content-negative)`,
rather than `var(--vtmn-semantic-color_content-danger)` (that doesn't exist - stating a `danger` variant right now output a black color)

- Yes

`danger` wording variant, if ever used...
```tsx
<VtmnIcon variant="danger" value="home-fill" />
```
...should be update to `negative`
```tsx
<VtmnIcon variant="negative" value="home-fill" />
```

❤️ Thank you!

### DIFF
--- a/packages/sources/react/src/components/VtmnIcon/VtmnIcon.tsx
+++ b/packages/sources/react/src/components/VtmnIcon/VtmnIcon.tsx
@@ -50,16 +50,14 @@ export const VtmnIcon: React.FC<VtmnIconProps> = ({
         return 'content-warning';
       case 'negative':
         return 'content-negative';
-      case 'neutral':
-        return '';
     }
   };
 
   const getIconColor = () => {
-    let iconColor = 'inherit';
+    let iconColor = undefined;
     if (color) {
       iconColor = `var(--vtmn-color_${color})`;
-    } else if (variant !== 'neutral') {
+    } else if (variant !== 'default') {
       iconColor = `var(--vtmn-semantic-color_${retrieveSemanticColor(
         variant,
       )})`;
--- a/packages/sources/react/src/components/VtmnIcon/VtmnIcon.tsx
+++ b/packages/sources/react/src/components/VtmnIcon/VtmnIcon.tsx
@@ -48,8 +48,8 @@ export const VtmnIcon: React.FC<VtmnIconProps> = ({
         return 'content-information';
       case 'warning':
         return 'content-warning';
-      case 'danger':
-        return 'content-danger';
+      case 'negative':
+        return 'content-negative';
       case 'neutral':
         return '';
     }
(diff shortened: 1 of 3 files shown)


packages/sources/react/src/components/VtmnIcon/VtmnIcon.tsx
```diff
@@ -48,19 +48,32 @@ export const VtmnIcon: React.FC<VtmnIconProps> = ({
...
+        return '';
     }
   };
 
+  const getIconColor = () => {
+    let iconColor = 'inherit';
+    if (color) {
+      iconColor = `var(--vtmn-color_${color})`;
+    } else if (variant !== 'neutral') {
+      iconColor = `var(--vtmn-semantic-color_${retrieveSemanticColor(
```
lauthieb, line 63: I understand your proposal with "neutral" to avoid `inherit` by default. But, when `variant` will be `neutral`, this code will output `var(--vtmn-semantic-color_)`, we need to find a better solution, please.

lauthieb (changes requested):
Thanks a lot @crotru for your proposal. Just one remark detailed in this PR.

packages/sources/react/src/components/VtmnIcon/VtmnIcon.tsx
lauthieb: Hello @crotru, do you have an idea about this?
Thx!

crotru: This actually works pretty fine.
The idea was to keep it black by default, but stating the color as inherit when a "neutral" variant is specified.

But...

Coming back to it, it actually makes more sense to keep the color as inherit (or not stated at all) when no variant is provided. 
I'll come up with something that way.

crotru:
Here is a new proposition without "neutral" variant.
icon color inherit from its parent whenever no variant is stated

packages/sources/react/src/components/VtmnIcon/VtmnIcon.tsx
lauthieb: Thanks a lot @crotru

lauthieb (approved):

lauthieb:
For me it's ok, thanks @crotru :tada:

PR #865, opened 2022-01-06
refactor(@vtmn/css-rating): improve accessibility with keyboard + screen reader

GaspardMathon:
## Description

- The previous version of the rating componet wasn't intuitive/accessible with keyboard + screen reader.
- The component has been reworked to be accessible

- No (non class name changes)
- Some aria props should be removed
- The class `vtmn-rating--disabled` can be removed and replaced by `aria-disabled="true"`

❤️ Thank you!


packages/showcases/css/stories/components/rating/examples/interactive.html
```diff
@@ -5,65 +5,166 @@
...
+      <input
+        type="radio"
+        name="rating2"
+        value="5"
+        id="rating2-5"
+        aria-label="5 stars out of 5"
+      />
+      <label for="rating2-5"></label>
     </div>
-    <span class="vtmn-rating_comment--primary" aria-label="Article rating"
```
lauthieb, line 62: Why did you delete aria labels for comments?

lauthieb (changes requested):
Thanks Gaspard!
+ you need to don't allow focus when disabled

lauthieb:
@GaspardMathon We also have a problem with `:hover` state:
Jan-10-2022 17-09-40
It works with values upper than the current value but not with before

packages/showcases/css/stories/components/rating/examples/interactive.html
GaspardMathon: I remobed this aria-label because according MDN : https://developer.mozilla.org/fr/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

Aria-label should be set up when the text is not visible. Here the text is visible so we shouldn't use it.

lauthieb:
as we still have issues in this PR (discussed together), I convert again to draft.

GaspardMathon:
> @GaspardMathon We also have a problem with `:hover` state: Jan-10-2022 17-09-40 It works with values upper than the current value but not with before.

Resolved in last commit

lauthieb (approved):

PR #948, opened 2022-01-26
feat(@vtmn/css): remove margin from input into label/helper-text

### LINKED ISSUE #946: bug: text input margin
**Describe the bug**

Right now text input margin is held by the `vtmn-text-input_container`.
Aim of this margin is - as far as I know - to give space between label / input / helper.

If the component doesn't hold a helper (or even a label), we get extra margin that we don't need :

Capture d’écran 2022-01-26 à 09 31 43

**Steps to reproduce**
    1. Go to 'https://decathlon.github.io/vitamin-web/@vtmn/showcase-css/?path=/story/components-forms-text-input--multiline'
    2. hover `#my-text-input-multiline-1` in inspector
    3. See useless bottom margin

**Expected behavior**
- keep outbound margin out of the component

**Browsers affected**

All

GaspardMathon:
## Changes description

- Move margins from the input to the label and the helper text to avoid useless margins when no label OR/AND helper text

## Context

- Text input has default margin even without label and text helper
- Close #946 

## Checklist

- No

### DIFF
--- a/packages/sources/css/src/components/forms/text-input/src/index.css
+++ b/packages/sources/css/src/components/forms/text-input/src/index.css
@@ -26,8 +26,6 @@
   font-weight: var(--vtmn-typo_font-weight--normal);
   font-size: rem(16px);
   line-height: 1;
-  margin-bottom: rem(4px);
-  margin-top: rem(4px);
   padding: rem(12px) rem(36px) rem(12px) rem(12px);
   color: var(--vtmn-semantic-color_content-primary);
   min-height: rem(48px);
@@ -65,6 +63,7 @@ textarea.vtmn-text-input {
   line-height: 1;
   display: block;
   width: fit-content;
+  margin-block-end: rem(4px);
 }
 
 .vtmn-text-input_container {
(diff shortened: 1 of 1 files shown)


packages/sources/css/src/components/forms/text-input/src/index.css
```diff
@@ -93,13 +92,12 @@ textarea.vtmn-text-input {
...
   font-size: rem(14px);
   line-height: 1;
-  margin-top: 0;
+  margin-block-start: rem(4px);
   width: fit-content;
 }
 
 .vtmn-text-input_helper-text--error {
   color: var(--vtmn-semantic-color_content-primary);
-  margin-top: rem(-2px);
```
crotru, line 102: It seems this was intended to keep height consistency between classic helper text (without icon) and alert ones (with icons)

By removing that negative value you will loose that height.

https://user-images.githubusercontent.com/22882255/151188877-15e96647-6144-4d85-a44e-b7ce077a5b06.mov

That can be kept though. By removing the top value of the alert icon and adding it some vertical-align

PR #955, opened 2022-02-03
feat(@vtmn/vue): add `VtmnIcon` component

ThomasLaigneauDecathlon:
## Changes description

Add Vue implementation of VtmnIcon component.

## Context

https://github.com/Decathlon/vitamin-web/issues/717

## Checklist

- No

### DIFF
--- a/packages/showcases/core/csf/guidelines/iconography.csf.js
+++ b/packages/showcases/core/csf/guidelines/iconography.csf.js
@@ -14,7 +14,7 @@ export const argTypes = {
     defaultValue: 24,
     control: {
       type: 'select',
-      options: [16, 24, 32, 64],
+      options: [16, 20, 24, 32, 64],
     },
   },
   color: {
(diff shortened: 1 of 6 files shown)


packages/showcases/core/csf/guidelines/iconography.csf.js
```diff
@@ -14,7 +14,7 @@ export const argTypes = {
     defaultValue: 24,
     control: {
       type: 'select',
-      options: [16, 24, 32, 64],
+      options: [16, 20, 24, 32, 64],
```
lauthieb, line 17: It is allowed in design?

lauthieb (changes requested):
Awesome :tada:
Just one remark

packages/showcases/core/csf/guidelines/iconography.csf.js
ThomasLaigneauDecathlon: Added it to match with the React component but it seems that only the 16, 24, and 32 pixels are available. It means we should remove the 20 and 64 size ? We should also make this update in the react component ?

GaspardMathon (changes requested):
See for changing the icon sizes that are not allowed (20, 64)

Issue #958, opened 2022-02-08
bug(@vtmn/svelte): components cannot be imported on server side


Tlahey:
Hello,

On svelte component (example VtmnButton.svelte), the component import the style on the `<script>` node.
This type of operation can cause side effects on the generated component on the SSR because is tires to interpret the css as script.

The quick solution is to move the css import into a `<style>` node.
To go further, during the build, inject the script directly into the component to make it totally autonomous (without node imports)

lauthieb:
Hi @Tlahey, thanks a lot for reporting this!
As this package is built by the community, we cannot help you more sorry.
Do you have contact that could help you resolve this issue?
@jleveugle @thollander if you have an idea. I thought we tested for SSR the Svelte components :/
Thanks a lot.
Laurent

PR #960, opened 2022-02-09
fix(@vtmn/svelte): move import style into <style> element

Tlahey:
Refer to #958 

On svelte component (example VtmnButton.svelte), the component import the style on the <script> node.
This type of operation can cause side effects on the generated component on the SSR because is tires to interpret the css as script.

The quick solution is to move the css import into a <style> node.
To go further, during the build, inject the script directly into the component to make it totally autonomous (without node imports)


GaspardMathon (approved):
Thanks for your contribution 💪

lauthieb (approved):
Thanks a lot @Tlahey :)

PR #967, opened 2022-02-11
feat(@vtmn/vue): add `VtmnToggle` and `VtmnCheckbox` reactivity

ThomasLaigneauDecathlon:
## Changes description

Update `VtmnToggle` and `VtmnCheckbox` components to make them ready to use in a Vue 3 project with reactivity.

Also add `VtmnCheckbox` standalone props to make it useable according to the design system. More explanation in Context section below.

## Context

I added a standalone props for `VtmnCheckbox` in order to be able to use it in all the possible context.

You only need this props when you want to display only one checkbox for example accepting or not the general condition in an ecommerce context. In this case we need our component to emit a true or false value and not the value that it would contain that would be pushed in an array when using multiple checkbox.

I know VTMN DS aim to release stateless component but in our case it's not possible to avoid this props and to emit the update:modelValue with the value when we need true or false based on checked state of input checkbox.

Feel free to reach me if you need more information or a live demo to better understand the trick.

## Checklist

- No

### DIFF
--- a/packages/sources/vue/src/components/selection-controls/VtmnCheckbox/VtmnCheckbox.vue
+++ b/packages/sources/vue/src/components/selection-controls/VtmnCheckbox/VtmnCheckbox.vue
@@ -1,34 +1,64 @@
 <script lang="ts">
-import '@vtmn/css-checkbox/dist/index-with-vars.css';
-import { defineComponent } from 'vue';
+import { defineComponent, PropType, reactive } from 'vue';
 
 export default /*#__PURE__*/ defineComponent({
   name: 'VtmnCheckbox',
   props: {
+    modelValue: {
+      type: [String, Number, Boolean, Array] as PropType<
+        string | number | boolean | []
+      >,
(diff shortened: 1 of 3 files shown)


packages/sources/vue/src/components/selection-controls/VtmnCheckbox/VtmnCheckbox.vue
```diff
@@ -1,48 +1,79 @@
 <script lang="ts">
-import '@vtmn/css-checkbox/dist/index-with-vars.css';
```
lauthieb, line 2: Why did you delete the CSS?

lauthieb (commented):
Thanks @ThomasLaigneauDecathlon 
Just a small change

lauthieb (changes requested):
Thanks @ThomasLaigneauDecathlon 
Just a small change

lauthieb (approved):

PR #1017, opened 2022-03-04
feat(@vtmn/svelte): add `VtmnChip` component

Tlahey:
Create a svelte `VtmnChip` component.

Features : 
- Add README.md on the root component `VtmnChip`
- Add component test thanks to jest :) 
- Add new libraries for testing on `packages/sources/svelte`
- Add task test on svelte build and project package.json
- Add more than 44 units tests on the `VtmnChip` component
- `VtmnChip` import component `VtmnBadge` for the filter badge and `VtmnButton` for the action variant


packages/sources/svelte/src/components/selection-controls/VtmnChip/VtmnChip.svelte
```diff
@@ -0,0 +1,99 @@
...
+      {disabled}
+    />
+  {/if}
+  {#if displayFilterBadge}
+    <VtmnBadge value={badgeValue} />
+  {/if}
+</div>
+
+<style lang="css">
+  @import '@vtmn/css-chips';
```
GaspardMathon, line 98: I think it is `@vtmn/css-chip`

GaspardMathon (changes requested):

packages/sources/svelte/src/components/selection-controls/VtmnChip/VtmnChip.svelte
Tlahey: Ho yes ! Sorry xD

Tlahey:
I have a question though.
Wouldn't it be better to make X components Chip for each variant. So, each one has the necessary props and would not share props not useful for the variant 🤔 

For example, only chip filter can have a badge, so if we create a `VtmnChipFilter` it's more clear to see props used by the component

In counterpart, we have more component to maintains, but it's more clear ... To discuss.

packages/sources/svelte/src/components/selection-controls/VtmnChip/enums.js
```diff
@@ -0,0 +1,11 @@
+export const VARIANTS = {
```
lauthieb, line 1: => `VTMN_CHIP_VARIANT`

Same for size :)

lauthieb (changes requested):
Thanks a lot @Tlahey, small changes :)

lauthieb:
+ please delete `package-lock.json` as we only use yarn.

lauthieb:
@Tlahey I understand your proposition but we need to be as near as possible from the documentation and Figma, that's why I prefer only one component, even if some variants cannot be present together... Thanks for your understanding

lauthieb (approved):
Thanks a lot @Tlahey :)

PR #1024, opened 2022-03-07
feat(@vtmn/svelte): add `VtmnModal` component

Tlahey:
Add a new component svelte `VtmnModal`

- Add tests on the component
- Workaround for a css issue (#1039)
- Add `animation` properties set at `true` by default. You can disable the show animation. We have some issues with this one on chrome.

### DIFF
--- a/packages/showcases/svelte/stories/components/overlays/VtmnModal/VtmnModal.stories.svelte
+++ b/packages/showcases/svelte/stories/components/overlays/VtmnModal/VtmnModal.stories.svelte
@@ -69,8 +69,19 @@
       in quia?
     </p>
     <div slot="actions" class="vtmn-modal_content_actions">
-      <VtmnButton variant="secondary">Go Back</VtmnButton>
-      <VtmnButton variant="primary">Yes, I understand</VtmnButton>
+      <VtmnButton
+        variant="secondary"
+        on:click={() => {
+          console.log('Go back');
+          show = false;
+        }}>Go Back</VtmnButton
+      >
+      <VtmnButton
+        variant="primary"
+        on:click={() => {
+          console.log('Yes, I understand'), (show = false);
+        }}>Yes, I understand</VtmnButton
+      >
     </div>
   </VtmnModal>
 </Story>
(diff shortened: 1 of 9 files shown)


Tlahey:
Need to tests if we can initialize the modal on the body for all the application

packages/showcases/svelte/stories/components/overlays/VtmnModal/VtmnModal.stories.svelte
```diff
@@ -0,0 +1,143 @@
...
+  import { VtmnModal, VtmnButton } from '@vtmn/svelte';
+  import { parameters } from '@vtmn/showcase-core/csf/components/overlays/modal.csf';
+  const argTypes = {
+    title: {
+      type: { name: 'string', required: true },
+      description: 'The printed title',
+      defaultValue: 'Modal title',
+      control: { type: 'text' },
+    },
+    visible: {
```
Tlahey, line 12: show instead visible

packages/sources/svelte/src/components/overlays/VtmnModal/VtmnModal.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Title of the modal
+   */
+  export let title;
+
+  /**
+   * Display or hide the modal
+   */
+  export let visible = false;
```
Tlahey, line 16: show instead visible

Tlahey:
So ... I do some changes on the component.
- Add `disableAnimation`, why ? See this one 
-- With class `show` on the modal 
image
-- Wihout class `show` on the modal
image

So, animation will infect the z-index.

Another very interesting addition only on svelte : 
- Add automatically a `<style>` to hide the overflow on body when the modal is open ! 🔥

lauthieb (changes requested):
Thanks a lot @Tlahey 

I have two remarks:
- It seems the control for showing the modal doesn't work on Storybook
- Can you please add "show = false" when we click on the buttons inside your example modal?

Thanks in advance :)

Tlahey:
> * when we click on the buttons inside your example modal?

I've tried to do some changes on the story, but args are only one way binding. Storybook don't handle to change this values on the documentation.

lauthieb (approved):
Awesome @Tlahey :)

PR #1037, opened 2022-03-11
feat(@vtmn/vue): disable inheritAttrs to components using attrs binding

ThomasLaigneauDecathlon:
## Changes description

Add inheritAttrs: false to components using v-bind="$attrs".

## Context

In Vue 3 we use v-bind="$attrs" to pass html attributes to a specific element of our component. We also need to add inheritAttrs: false to avoid passing those html attributes to the first html element inside the Vue component. By default Vue pass attributes to the first HTML element found inside the <template> tag of the components. When we use v-bind="$attrs" withtout inheritAttrs: false the attributes are passed to the first html element and also the targeted element with  v-bind="$attrs" .

## Checklist

- No

More info here : https://vuejs.org/guide/components/attrs.html#disabling-attribute-inheritance


lauthieb (approved):
Thanks @ThomasLaigneauDecathlon and I trust you for no breaking change :)

lauthieb (changes requested):
As discussed, we wait for checking our consumers because there could be a little breaking change.

lauthieb (approved):

PR #1043, opened 2022-03-14
feat(@vtmn/svelte): add `VtmnAlert` component

Tlahey:
Create a new `VtmnAlert` component for svelte

How it works :

- VtmnAlert need to be put on the main body of the application.
- VtmnAlertItem are used by the VtmnAlert to display toasts
- A store vtmnAlertStore can be imported everywhere on the application, you can use the function `vtmnAlertStore.send({ title: 'Hello world!' })` to send alert to display

```svelte
vtmnAlertStore.send({
        'title': 'This is the title of the alert',
        'description': 'Alert are used to draw the users attention to an important information',
        'withCloseButton': true,
        'variant': 'info',
        'aria-labelledby': 'Storybook',
        'aria-describedby': args.variant,
      });
```

The component wait 8 seconds before display the next alert and so one ...
We need to up all elements on an array. The trick here is to take the first element of the array and use the id for the key.
If we don't do that, the same div are use for each key and all alert share the same timeout.

image


Margot-ElMadi (approved):

lauthieb (approved):
Perfect @Tlahey, thanks a lot!

PR #1063, opened 2022-03-23
fix(@vtmn/react): bad props type given to DOM in `VtmnTextInput`

wlalele:
## Context

We pass all the component props to the DOM as we receive it.
Some of them are in boolean type which causes a console warning.
(See the screenshots bellow)

## Changes description

We simply cast all the props as string before passing them to the DOM.

## Checklist

No

## Other information

Capture d’écran 2022-03-23 à 10 50 36

### DIFF
--- a/packages/sources/react/src/components/forms/VtmnTextInput/VtmnTextInput.tsx
+++ b/packages/sources/react/src/components/forms/VtmnTextInput/VtmnTextInput.tsx
@@ -90,6 +90,9 @@ export const VtmnTextInput = ({
   onIconClick,
   ...props
 }: VtmnTextInputProps) => {
+  const propsToString = Object.keys(props).map((propKey) =>
+    props[propKey].toString(),
+  );
   return (
     <>
       {labelText && (
@@ -110,7 +113,7 @@ export const VtmnTextInput = ({
           aria-describedby={
             (helperText && `${identifier}-helper-text`) || undefined
           }
-          {...props}
+          {...propsToString}
         />
       ) : (
         <div className="vtmn-text-input_container">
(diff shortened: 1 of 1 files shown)


crotru (approved):

packages/sources/react/src/components/forms/VtmnTextInput/VtmnTextInput.tsx
```diff
@@ -90,6 +90,9 @@ export const VtmnTextInput = ({
   onIconClick,
   ...props
 }: VtmnTextInputProps) => {
+  const propsToString = Object.keys(props).map((propKey) =>
+    props[propKey].toString(),
+  );
```
lauthieb, line 95: Thanks! Do you think this could be more global for all components? (I think about making an High Order Component for example or a function that we use in all React components).

lauthieb:
@wlalele @crotru are you sure there will not have any breaking change for all our current consumers of `@vtmn/react` and its `VtmnTextInput` component, please?

packages/sources/react/src/components/forms/VtmnTextInput/VtmnTextInput.tsx
wlalele: I don't think a HOC would be a good solution in that case but a function would definitively do the trick.
(Function that will do the same thing, cast props to string)
But we would indeed need to edit every react component to use it.

wlalele:
> @wlalele @crotru are you sure there will not have any breaking change for all our current consumers of `@vtmn/react` and its `VtmnTextInput` component, please?

100% sure

packages/sources/react/src/components/forms/VtmnTextInput/VtmnTextInput.tsx
lauthieb: @wlalele okay for me, can you please add this fix for all components and create a function for this in a `utils` folder like we did in Svelte stack? Thanks a lot.?

wlalele: @lauthieb this bug fix is related to the text input component, I suggest you merge this one and I'll open a new PR for the other components as they require an in depth testing before merging.

lauthieb (approved):

Issue #824, opened 2021-12-28
feat: add quantity component


lauthieb:
**Feature request**

Add quantity component

**Documentation links**

Documentation: https://www.decathlon.design/726f8c765/p/078570-list-beta/b/9115e7
Figma: https://www.figma.com/file/zDZIyayUlr1yTWrsi7cFoo/Vitamin---Web-UI-kit-(v0)?node-id=6937%3A16423

**Checklist**

- [x] CSS (thanks @GaspardMathon)
- [x] React (thanks @0x2A-git)
- [x] Svelte (thanks @Tlahey)
- [x] Vue (thanks @GaspardMathon)

PR #1097, opened 2022-03-30
feat(@vtmn/react): add `VtmnQuantity` component

0x2A-git:
## Changes description

This P.R aims to add the quantity component

## Context

#824

## Checklist

- No

### DIFF
--- a/packages/showcases/core/csf/components/selection-controls/quantity.csf.js
+++ b/packages/showcases/core/csf/components/selection-controls/quantity.csf.js
@@ -17,14 +17,6 @@ export const parameters = {
 };
 
 export const argTypes = {
-  onQuantityChange: {
-    type: { name: 'string', required: false },
-    description: 'Called when value has changed',
-    control: {
-      type: 'function',
-    },
-  },
-
   id: {
     type: { name: 'string', required: true },
     description: 'Id of the input.',
--- a/packages/showcases/core/csf/components/selection-controls/quantity.csf.js
+++ b/packages/showcases/core/csf/components/selection-controls/quantity.csf.js
@@ -17,6 +17,14 @@ export const parameters = {
 };
 
 export const argTypes = {
+  onQuantityChange: {
+    type: { name: 'string', required: false },
+    description: 'Called when value has changed',
+    control: {
+      type: 'function',
+    },
+  },
+
   id: {
     type: { name: 'string', required: true },
     description: 'Id of the input.',
--- a/packages/showcases/react/stories/components/selection-controls/VtmnQuantity/VtmnQuantity.stories.tsx
+++ b/packages/showcases/react/stories/components/selection-controls/VtmnQuantity/VtmnQuantity.stories.tsx
@@ -9,7 +9,16 @@ import {
 export default {
   title: 'Components / Selection controls / VtmnQuantity',
   component: VtmnQuantity,
-  argTypes: argTypes,
+  argTypes: {
+    onChange: {
+      type: { name: 'string', required: false },
+      description: 'Called when value has changed',
+      control: {
+        type: 'function',
+      },
+    },
+    ...argTypes,
+  },
   parameters,
 } as Meta;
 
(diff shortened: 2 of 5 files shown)


packages/showcases/core/csf/components/selection-controls/quantity.csf.js
```diff
@@ -17,6 +17,14 @@ export const parameters = {
...
 
 export const argTypes = {
+  onQuantityChange: {
+    type: { name: 'string', required: false },
+    description: 'Called when value has changed',
+    control: {
+      type: 'function',
+    },
+  },
+
```
lauthieb, line 27: If you add this here, it will be add in `@vtmn/svelte` showcase also.
What's the use case, please?

@Tlahey @GaspardMathon for your information, do you think this could be interesting? In `@vtmn/svelte` they have chosen to dispatch events "subtract" and "add" each time we click to their respective buttons:

https://github.com/Decathlon/vitamin-web/blob/f240375daf7b89e5698a1867a479c169450353a0/packages/sources/svelte/src/components/selection-controls/VtmnQuantity/VtmnQuantity.svelte#L54-L61

If yes, I would have prefered `onChange` because `onQuantityChange` is too long as we are already in a `<VtmnQuantity>` component.

lauthieb (changes requested):
Thanks @0x2A-git for this contribution again!
Just a remark :)

packages/showcases/core/csf/components/selection-controls/quantity.csf.js
0x2A-git: Hi Laurent,

This callback is used to retrieve the latest quantity's value when updated. 

As for the substract and add methods it could be a great idea if we could find a way to make them work on react when clicking the buttons ( At first I tried something like that but didn't manage to find a way to retrieve the updated value, maybe we could do some checks in the useEffect hook and call the approriate callback based on these checks ? )

Tlahey: We have made 2 differents calls to split the action.
But in fact, it is possible in svelte to have just 1 dispatch with a parameter "add" or "subtract" which it defined the action.

lauthieb: Good idea @Tlahey :)
So @0x2A-git I propose you to keep this but just renamed it to `onChange` instead of `onQuantityChange`.

Tlahey: But `onChange` should be specific to react property.
On svelte, it is not a prop. Please don't add it on `[quantity.csf.js](https://github.com/Decathlon/vitamin-web/pull/1097/files/93b2b5adc8e023831ca77f3428c3f8fbe0997119#diff-cbe5310d0c26ff3cd5f4d9060391727ae82e3e2078854fcf2060ce81224351fd)`

GaspardMathon (approved):

packages/showcases/core/csf/components/selection-controls/quantity.csf.js
lauthieb: Yes, we will take care of this @Tlahey ;)

Tlahey: @0x2A-git good choice for the edit, on svelte we use `on:blur` action we need to change it !

lauthieb: @0x2A-git Thanks a lot! Is it possible to use also onBlur on React part as Svelte did?

0x2A-git: > @0x2A-git Thanks a lot! Is it possible to use also onBlur on React part as Svelte did?

It's done, I've forwarded a ref to the input :)

lauthieb: @0x2A-git thanks Lucas!

PR #1117, opened 2022-04-07
fix(@vtmn/svelte): custom class should be apply on the `VtmnModal` content

Tlahey:
When we use the `VtmnModal` and want to apply custom style on it, the style are applied on the main container `vtmn-modal` and not the content of the modal `vtmn-modal_content`.

The first level only contains the overlay and the content.

In fact, if we want to apply a custom height to the modal for example : 

Without the fix
```
<style>
.variant-full-screen {
  .vtmn-modal-content {
    height: 100vh;
  }
}
</style>

<VtmnModal class="modal-full-screen" />
```

With the fix 
```
<VtmnModal class="vtmn-h-screen" />
```

### DIFF
--- a/packages/sources/svelte/src/components/overlays/VtmnModal/VtmnModal.svelte
+++ b/packages/sources/svelte/src/components/overlays/VtmnModal/VtmnModal.svelte
@@ -17,7 +17,7 @@
    */
   export { className as class };
 
-  $: componentClass = cn('vtmn-modal', className);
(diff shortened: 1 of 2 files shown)


lauthieb (approved):

lauthieb:
Hi @Tlahey okay I understand!
@GaspardMathon to have in mind for other implementations in Vue & React 👌🏻

PR #1138, opened 2022-04-28
fix(@vtmn/css): edit `button` icon sizes, increase icon sizes for `icon-alone` variant, a11y 

GaspardMathon:
## Changes description

For icon-left / icon-right vairants :

- Small button icon goes from `--vtmn-typo_text-3-font-size` to `16px`
- Medium (default) button icon goes from `--vtmn-typo_text-2-font-size` to `20px`
- Small button icon goes from `--vtmn-typo_title-5-font-size` to `24px`

For icon-alone vairants :

- Small button icon goes from `--vtmn-typo_text-3-font-size` to `20px`
- Medium (default) button icon goes from `--vtmn-typo_text-2-font-size` to `24px`
- Small button icon goes from `--vtmn-typo_title-5-font-size` to `32px`

Accessibility:

- `width` turned into `inline-size`
- `height` turned into `block-size` 

## Context

- Icon in `buttons` were smaller than in the design library.
- Add a11y on the css sizes

## Checklist

- No

### DIFF
--- a/packages/sources/css/src/components/actions/button/src/index.css
+++ b/packages/sources/css/src/components/actions/button/src/index.css
@@ -317,7 +317,7 @@
 
 .vtmn-btn_size--large.vtmn-btn--icon-right span[class^='vtmx-'] {
   padding-inline-start: rem(12px);
-  rem(24px)
+  font-size: rem(24px);
 }
 
 .vtmn-btn_size--large.vtmn-btn--icon-left svg {
--- a/packages/sources/css/src/components/actions/button/src/index.css
+++ b/packages/sources/css/src/components/actions/button/src/index.css
@@ -239,7 +239,7 @@
 }
 
 .vtmn-btn--icon-alone span[class^='vtmx-'] {
-  font-size: var(--vtmn-typo_text-2-font-size);
+  font-size: var(--vtmn-typo_title-5-font-size);
 }
 
 .vtmn-btn_size--small.vtmn-btn--icon-alone {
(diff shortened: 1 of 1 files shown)


packages/sources/css/src/components/actions/button/src/index.css
```diff
@@ -259,7 +259,7 @@
 }
 
 .vtmn-btn_size--medium.vtmn-btn--icon-alone span[class^='vtmx-'] {
-  font-size: var(--vtmn-typo_text-2-font-size);
+  font-size: var(--vtmn-typo_title-5-font-size);
```
lauthieb, line 262: Semantically, it's weird to use a "title" design token in a button.
Can you check with @N0t-Simon, Sabrina or Marie? Thanks.

lauthieb (changes requested):

packages/sources/css/src/components/actions/button/src/index.css
GaspardMathon: Semantically, I agree with you but this is the token for the wanted size. Otherwise I would have to set the size in px.

GaspardMathon: PS: it is for the icon, not the label text

Sabrinavigil: Yes, it's for icons (the button "icon-alone"). 
I will let you decide on the naming convention of the token; is it a new one? 
In any case, it needs to be larger than the "classic body text."

GaspardMathon: @lauthieb Do you mean that we should create specific design tokens for icon sizes ?

GaspardMathon: Ok, problem solved I think. As the icons must not change size according to the device screen size. We have to set icon size in px, not use typography design token.

GaspardMathon:
Waiting for the design part before merging

Tlahey (approved):

lauthieb (approved):

PR #1165, opened 2022-05-17
fix(@vtmn/css): fix scroll padding mobile

Tlahey:
On small device, the scroll bar are under content.
But in some cases, this solution produce side effects.

| Before | After  |
|--------|--------|
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-05-17 at 10 05 05  | Simulator Screen Shot - iPhone 13 Pro Max - 2022-05-17 at 10 06 07 |

The solution is to export the padding to the mobile component so that the scroll is outside the container


lauthieb (approved):
Perfect! Thanks @Tlahey :)

PR #1214, opened 2022-07-26
build(@vtmn/svelte): inject CSS to `dist` folder by switching to PostCSS

Tlahey:
Change the rollup configuration.
Switch to postcss to insert the `@import`ed css into the dist folder.

This enhancement doesn't required anymore the `@vtmn/css-*` as dependencies.

In the dist output with this changes : 

image

- Add `rollup-plugin-terser` in order to minify the bundle
- Move `rollup-plugin-css-only` to `rollup-plugin-postcss`

### DIFF
--- a/packages/sources/svelte/package.json
+++ b/packages/sources/svelte/package.json
@@ -31,7 +31,10 @@
     "test": "jest src",
     "test:watch": "npm run test -- --watch"
   },
-  "dependencies": {
+  "devDependencies": {
+    "@rollup/plugin-node-resolve": "^13.1.2",
+    "@testing-library/jest-dom": "^5.16.2",
+    "@testing-library/svelte": "^3.0.3",
     "@vtmn/css-accordion": "^0.5.0",
     "@vtmn/css-alert": "^0.8.3",
     "@vtmn/css-badge": "^0.13.0",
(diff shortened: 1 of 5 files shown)


Tlahey:
We need a fix from svelte to merged this PR in the futur.
See : 
- https://github.com/sveltejs/svelte/pull/6646
- https://github.com/sveltejs/svelte/issues/6584

packages/sources/svelte/package.json
```diff
@@ -64,20 +67,18 @@
     "@vtmn/css-toast": "^0.7.2",
     "@vtmn/css-toggle": "^0.8.2",
     "@vtmn/css-tooltip": "^0.8.2",
-    "@vtmn/icons": "^0.14.0"
-  },
-  "devDependencies": {
-    "@rollup/plugin-node-resolve": "^13.1.2",
-    "@testing-library/jest-dom": "^5.16.2",
-    "@testing-library/svelte": "^3.0.3",
+    "@vtmn/icons": "^0.14.0",
```
AxelPeter, line 70: We could leave the `@vtmn` packages as dependencies, it could be a breaking changes if the `/src` folder is still used.

packages/sources/svelte/package.json
```diff
@@ -109,7 +110,8 @@
       "url": "http://www.apache.org/licenses/LICENSE-2.0"
     }
   ],
-  "main": "src/index.js",
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
   "svelte": "src/index.js",
```
AxelPeter, line 115: No change here ?

Tlahey: Nothing change here. In fact, instead of using the component not compiled, we used the compiled one.
Before, we have to use the src index because the dist doesn't expose an index.js who export all components as js.

AxelPeter: I mean, why not `"svelte": "dist/index.js",` like the others entries.
Like this, we always use by default the `/dist` folder.

packages/sources/svelte/package.json
Tlahey: Not necessary because the dist deliver all the needed sources.
In fact, when we compile the source code, now, we getting the sources from the dist folder.

But yes, of course, it can be a breaking change if we use again the `/src` folder

packages/sources/svelte/package.json
Tlahey: Yes, I will check that, but if I do that, I need to reimport the `@vtmn` as dependances ^^

packages/sources/svelte/package.json
AxelPeter: If the `/src` folder is still available and accessible, we should leave them as dependencies.

Like said by NPM:
```
Development dependencies are intended as development-only packages, that are unneeded in production. For example testing packages, webpack or Babel.
```
https://nodejs.dev/en/learn/npm-dependencies-and-devdependencies
https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

AxelPeter (dismissed):

AxelPeter (approved):

Tlahey:
@lauthieb it's ok for us, we can push it :)

lauthieb (approved):

lauthieb (approved):

lauthieb:
Thanks a lot @Tlahey & @AxelPeter for this great work!

PR #1220, opened 2022-07-26
fix(@vtmn/svelte): `VtmnDivider` component `labelId` & default `<slot></slot>`

lauthieb:
## Changes description

This PR fixes two problems on `VtmnDivider` Svelte component.
1. Divider has an empty span on several showcases like this one in `VtmnTabs` _(thanks Christophe Drag for reporting me this)_: https://decathlon.github.io/vitamin-web/@vtmn/showcase-svelte/?path=/story/components-navigation-vtmntabs--overview
CleanShot 2022-07-26 at 21 49 31@2x
2. There is warning when `yarn test` for the `VtmnDivider` component because `labelId` is undefined. That's why I propose to use the `uuid()` function to generate a unique identifier if not present (for a11y purposes).

## Checklist

- No

### DIFF
--- a/packages/sources/svelte/src/components/structure/VtmnDivider/VtmnDivider.svelte
+++ b/packages/sources/svelte/src/components/structure/VtmnDivider/VtmnDivider.svelte
@@ -1,5 +1,4 @@
 <script>
-  import { onMount } from 'svelte';
   import { cn } from '../../../utils/classnames';
   import { uuid } from '../../../utils/math';
 
@@ -21,7 +20,7 @@
   /**
    * @type {string} Id of the label
    */
-  export let labelId = undefined;
+  export let labelId = uuid();
 
   let className = undefined;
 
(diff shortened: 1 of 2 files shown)


packages/sources/svelte/src/components/structure/VtmnDivider/VtmnDivider.svelte
```diff
@@ -28,6 +30,12 @@
    */
   export { className as class };
 
+  onMount(async () => {
```
Tlahey, line 33: You can set the `uuid` by default on the export let 

```
export let labelId = uuid();
```

If the user define the `labelId`, it takes the defined one :)

packages/sources/svelte/src/components/structure/VtmnDivider/VtmnDivider.svelte
```diff
@@ -43,7 +51,7 @@
   aria-labelledby={labelId}
   {...$$restProps}
 >
-  {#if $$restProps['slot'] !== ''}
+  {#if $$restProps['slot'] !== undefined && $$restProps['slot'] !== ''}
```
Tlahey, line 54: Maybe use just ?
```
{#if $$restProps['slot']} 
```

PR #1218, opened 2022-07-26
fix(@vtmn/css-tailwind-preset): add types

astahmer:
## Changes description
emit ts declaration files to get full typesafety on @vtmn/css-tailwind-preset

## Context
https://github.com/Decathlon/vitamin-web/issues/1215#issuecomment-1195333961

## Does this introduce a breaking change?
- No

## Other information

i'm not completely familiar with the repo and its multiple packages so I'd like if you could double check that I didnt break anything

PR #1222, opened 2022-07-26
fix(@vtmn/css-tailwind-preset): add types

astahmer:
https://github.com/Decathlon/vitamin-web/pull/1218

don't merge unless everything passes please ! would appreciate if you could double-check since i'm not familiar enough with the repo

### DIFF
--- a/packages/sources/css/presets/tailwind/package.json
+++ b/packages/sources/css/presets/tailwind/package.json
@@ -33,11 +33,6 @@
   "devDependencies": {
     "typescript": "^4.7.4"
   },
-  "dependencies": {
-    "@vtmn/css": "^0.84.1",
-    "@vtmn/css-design-tokens": "^0.20.0",
-    "@vtmn/css-utilities": "^0.4.2"
-  },
   "peerDependencies": {
     "tailwindcss": ">=1.9.0 || ^2.0.0 || ^3.0.0"
   },
--- a/packages/sources/css/presets/tailwind/package.json
+++ b/packages/sources/css/presets/tailwind/package.json
@@ -19,11 +19,20 @@
     "url": "git+https://github.com/Decathlon/vitamin-web.git"
   },
   "license": "Apache-2.0",
-  "main": "./index.js",
+  "main": "./dist/index.js",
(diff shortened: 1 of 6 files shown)


packages/sources/css/presets/tailwind/package.json
```diff
@@ -19,15 +19,19 @@
...
-    "theme.js"
+    "src",
+    "dist",
+    "package.json",
+    "tsconfig.json"
   ],
-  "dependencies": {
-    "@vtmn/css": "^0.84.1",
-    "@vtmn/css-design-tokens": "^0.20.0",
-    "@vtmn/css-utilities": "^0.4.2"
```
lauthieb, line 30: It resolves the circular dependency yes.
But it's a breaking change for users that didn't think to download these packages when consuming our preset.
Before, they came naturally so they can add what we said in our `README` of this preset, and with your proposal, it will break their imports because these dependencies will not come naturally.
Maybe we can add them in `peerDependencies` but I think nx will still say that it's a circular problem.
The only thing that I think it should work is to do this in `"prebuild"` script. By modifying also the root `package.json` file in order to launch `"prebuild"` script in all packages.

lauthieb (changes requested):
Yes, sorry again Alexandre for merging this quickly. Just one remark below.
Thanks for your help.

packages/sources/css/presets/tailwind/package.json
astahmer: you definitely have more context than me there so I'll let you do the required changes regarding the circular deps as I do not have enough knowledge not to break things 👍

lauthieb (approved):

lauthieb (approved):

lauthieb:
Thanks again @astahmer.

I've made my changes in this commit: https://github.com/Decathlon/vitamin-web/pull/1222/commits/d84b175453799ba2a212547145348fa230f2ef43

As the CI is green:
CleanShot 2022-07-27 at 21 30 14@2x

I will merge it, there should be no problems.

Again, sorry for the problem with the old PR that I had to revert...

Have a nice evening,
Laurent

PR #1224, opened 2022-07-29
fix(@vtmn/svelte): prevent body scroll on mobile for `VtmnModal`

Tlahey:
When `VtmnModal` are open, we apply a `overflow: hidden` on the body.
In fact, this solution works pretty well on desktop, but on mobile devices the touch are always enable and in some cases, the user can scroll over the body in the background.

To prevent this effect, we apply a `touch-action:none` on the body when the modal are open.


lauthieb (approved):

lauthieb:
Thanks @Tlahey :)

PR #1233, opened 2022-08-05
refactor(@vtmn/css): `tabs` demo showcase + rework a11y

GaspardMathon:
## Changes description

- Add behavior demo script for accessibility conformity
- Add tabpanel in the showcases to improve accessiblity purposes
- Add missing ARIA attributes
- Remove the `nav` tag as it was no necessary
- Replace the `a` by a `button`
- Update CSS in consequence (no more `.selected`)

## Context

- The `tabs` component wasn't accessible, now you can move across the different tabs with the left and right arrow keys. The tabs focus on the next element after the nav.

## Checklist

Yes : 
- remove the `nav` tag and move the class from it to the `ul` tag.
- remove all occurrences of `class='selected'`
- replace `a` with `button`, remove the `href` attribute.


lauthieb (changes requested):
thanks @GaspardMathon, sorry but there's a problem when we load directly the demo for the first time. When I "tab", the focus is displayed one out of two times. When I change to another story and come back, it works.

https://user-images.githubusercontent.com/9600228/183220391-1fd8d10f-6e1d-4f93-b865-8e28bde9673b.mp4

GaspardMathon:
Hi @lauthieb,[ I followed this link ](https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic.html)to set up the tabs keyboard component (I forgot the 'Home' and 'End' keys though I will add that). Once you "tab" on one tabList item, you must use arrow keys to navigate left and right. If the "tab" key is pressed a second time, the focus is moved on the next element following the tab component.

GaspardMathon:
Hi @lauthieb and @thibault-mahe, I have updated the code. I can't find a solution for the tab behavior. Currently, when you tab it will blur the last tab element (and the first one for tab + shift key). Feel free to contribute to this PR if you find a solution.

lauthieb (approved):

lauthieb:
Thanks @GaspardMathon 
@thibault-mahe I let you check also :)

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
```diff
@@ -1,9 +1,8 @@
 <div class="block">
   <nav id="demo-tabs-nav" aria-label="Tabs list" class="vtmn-tabs">
     <ul id="demo-tabs-tablist" role="tablist">
-      <span id="block-focus-trap-start" tabindex="-1"></span>
       <li id="tab-demo-1" role="tab" aria-selected="false">
-        <a href="" class="selected">Tab Name</a>
+        <a href="">Tab Name</a>
```
thibault-mahe, line 5: Not a fan of empty href, shouldn't we use `button` tags or set the `tabpanel` id here ?

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
```diff
@@ -14,8 +13,63 @@
...
-      <span id="block-focus-trap-end" tabindex="-1"></span>
     </ul>
     <div class="vtmn-divider" role="separator"></div>
   </nav>
 </div>
+
+<script>
+  let listTabs = document.querySelectorAll('#demo-tabs-tablist li > a');
+  let nav = document.getElementById('demo-tabs-nav');
+
```
thibault-mahe, line 24: Could be `const` ?
IMHO, regarding accessibility setting and testing, I often prefer to use accessible selectors to be sure role and aria attributes are set, for instance : 
```
const tabList = document.querySelector('[role=tablist]#demo-tabs-tablist');
```

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
```diff
@@ -14,8 +13,63 @@
...
+  listTabs.forEach((child, index, parent) => {
+    child.addEventListener('keydown', (e) => {
```
thibault-mahe, line 26: Again, just my personal opinion here, but also not a fan of single letter variable :) (even for e, but could be a specific case as it is widely used and understood)

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
```diff
@@ -14,8 +13,63 @@
...
+          e.preventDefault();
+          if (e.shiftKey) {
+            console.log('shift');
+            listTabs[0].focus();
+            listTabs[0].blur();
+          } else {
+            console.log('no shift');
+            listTabs[listTabs.length - 1].focus();
+            listTabs[listTabs.length - 1].blur();
+          }
```
thibault-mahe, line 58: According to the[ W3C accessibility tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/) you shared: 

> _When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable._

So arrows should be used to navigate the tablist, but tab key should be used to enter inside the tabpanel.

To do that you could set all tablist elements (tab) to a tabindex of -1, except for the active tab which should remain focusable. This way the next focusable element would be the first visible tabpanel.

I'll share with you a demo if you want

thibault-mahe (changes requested):
I think I don't have all the context elements here, I don't understand why is there no tabpanels on this demo 😬 Don't we have to check the whole component accessibility, meaning also the interaction between the tabs and their panels ?

If it is ok for you i'll commit another demo with the tabpanels so we can discuss this subject  🙂

thibault-mahe:
@GaspardMathon @lauthieb I commited some code for a new demo to discuss, with some fixes and tabpanels.
I'm available to discuss this, but feel free to revert whole commit if I don't have all context elements 🙂

lauthieb:
Thanks @GaspardMathon for this work.
I agree with @thibault-mahe and his suggestions regarding the patterns that we need to respect even if they will have some breaking changes for consumers, we can manage this transition as we don't have a lot of consumers for this component (< 10 apps).
I let you check together, feel free to reach me if you need further help.

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
GaspardMathon: Totally agree with you, GitHub Copilot is not perfect yet ^^'

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
GaspardMathon: In our case we don't have a tabpanel associated with the tabs. Do you think we should adda tabpanel ? In this case we would need a design template for that as it is not included in the basic composant

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
```diff
@@ -1,21 +1,196 @@
 <div class="block">
-  <nav id="demo-tabs-nav" aria-label="Tabs list" class="vtmn-tabs">
-    <ul id="demo-tabs-tablist" role="tablist">
-      <span id="block-focus-trap-start" tabindex="-1"></span>
-      <li id="tab-demo-1" role="tab" aria-selected="false">
-        <a href="" class="selected">Tab Name</a>
+  <nav id="demo-tabs-nav" class="vtmn-tabs">
+    <ul id="demo-tabs-tablist" role="tablist" aria-label="Tabs navigation">
```
thibault-mahe, line 3: @GaspardMathon `ul` and `li` being only semantic tags to introduce a list, isn't better to let the `aria-label="Tabs navigation"` on the `nav` tag ? It is the element that the user will reach with tab and screen-reader and thus the element that needs to be announced with an aria-label no ?

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
```diff
@@ -1,21 +1,196 @@
...
+      id="tabpanel-demo-1"
+      aria-labelledby="tab-demo-1"
+      tabindex="0"
+    >
+      Tab Name 1 <br />
+      Lorem, ipsum dolor sit amet consectetur adipisicing elit. Velit tenetur
+      voluptatum asperiores quis eum perspiciatis repellat temporibus ab aperiam
+      cumque deserunt eveniet dolore excepturi, error veniam quidem ipsum
+      necessitatibus? Sequi?
+    </section>
```
thibault-mahe, line 58: @GaspardMathon The text is a single node, what's the reason to put it on multiple lines? It makes it harder to manage no?

thibault-mahe (approved):

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
7studio: To go further, you should not use the `<nav>` HTML element here. 

The pattern to follow for this component is [Tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/). 

The `<nav>` HTML element must be used only in some specific cases : [Menu et barre de navigation (fr)](https://www.numerique.gouv.fr/publications/rgaa-accessibilite/methode-rgaa/glossaire/#menu-et-barre-de-navigation)

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
```diff
@@ -1,21 +1,196 @@
 <div class="block">
-  <nav id="demo-tabs-nav" aria-label="Tabs list" class="vtmn-tabs">
-    <ul id="demo-tabs-tablist" role="tablist">
-      <span id="block-focus-trap-start" tabindex="-1"></span>
-      <li id="tab-demo-1" role="tab" aria-selected="false">
-        <a href="" class="selected">Tab Name</a>
+  <nav id="demo-tabs-nav" class="vtmn-tabs">
+    <ul id="demo-tabs-tablist" role="tablist" aria-label="Tabs navigation">
+      <li>
+        <a
```
7studio, line 5: Please use `<button>` instead of `<a>` :pray:

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
GaspardMathon: You're right, it is not needed for such a component. I will remove it. Thanks for your awareness.

packages/showcases/css/stories/components/navigation/tabs/examples/demo.html
GaspardMathon: I think the prettier config automatically set long texts on several lines. @thibault-mahe

packages/sources/css/src/components/navigation/tabs/src/index.css
```diff
@@ -9,24 +9,21 @@
...
+.vtmn-tabs > li {
   display: inline-flex;
   padding-inline: rem(2px);
   padding-block-start: rem(2px);
 }
 
-.vtmn-tabs a,
-.vtmn-tabs_size--medium a {
+.vtmn-tabs > li > button,
+.vtmn-tabs_size--medium a > li > button {
```
7studio, line 26: I think you forgot to remove the `a` in the selector.

packages/sources/svelte/src/components/navigation/VtmnTabsItem/test/VtmnTabsItem.spec.js
```diff
@@ -2,13 +2,6 @@ import { render } from '@testing-library/svelte';
 import VtmnTabsItem from './../VtmnTabsItem.svelte';
 
 describe('VtmnTabsItem', () => {
-  test('Should be have a href', () => {
-    const { container } = render(VtmnTabsItem, { href: '#test' });
-    expect(container.getElementsByTagName('a')[0].getAttribute('href')).toEqual(
-      '#test',
-    );
-  });
-
```
lauthieb, line 11: Could be nice to keep this unit test but replace it with "Should have a button"

packages/sources/react/src/components/navigation/VtmnTabs/VtmnTabsItem.tsx
```diff
@@ -24,44 +18,28 @@ export interface VtmnTabsItemProps
...
-      onClick();
-    }
-  };
-
+export const VtmnTabsItem = React.forwardRef<
+  HTMLButtonElement,
+  VtmnTabsItemProps
+>(({ icon, badgeValue, children, ...props }, ref) => {
   return (
     <li role="tab">
```
lauthieb, line 33: There's a inconsistency here. In Svelte & Vue implementations, you put `role="tab"` on the `<button>`, not `<li>`.

lauthieb (changes requested):
Thanks @GaspardMathon, just few changes

packages/showcases/css/stories/components/navigation/tabs/examples/medium.html
```diff
@@ -1,110 +1,174 @@
...
+    role="tablist"
+    aria-label="Medium tabs navigation"
+  >
+    <li>
+      <button role="tab" aria-controls="tabpanel-medium-1" aria-selected="true">
+        Tab Name 1
+      </button>
+    </li>
+    <li>
+      <button role="tab">Tab Name 2</button>
```
thibault-mahe, line 13: Pure html showcase is a bit odd because none of the element here seems to be reactively/dynamically managed. I mean, in order to make this demo WCAG compliant, each button should have an `id` (to be paired with the `aria-labelledby` of the corresponding panel) and an `aria-selected` attribute defaulting to `false`. Not selected button should also have a `tabindex` set to `-1` so the next focusable element is the panel itself. Should we adapt the demo in this direction ? @GaspardMathon @lauthieb 

We could discuss that on videoconference if you want

packages/showcases/css/stories/components/navigation/tabs/examples/medium.html
```diff
@@ -1,110 +1,174 @@
...
+      <button role="tab">Tab Name 4</button>
+    </li>
+  </ul>
+  <div class="vtmn-divider" role="separator"></div>
+  <div class="vtmn-w-full vtmn-p-5 vtmn-bg-background-primary">
+    <section
+      role="tabpanel"
+      id="tabpanel-medium-1"
+      aria-labelledby="tabpanel-medium-1"
+      tabindex="0"
```
thibault-mahe, line 28: I'm not sure if this is necessary

packages/showcases/svelte/stories/components/navigation/VtmnTabs/VtmnTabs.stories.svelte
```diff
@@ -26,7 +26,7 @@
       {#each tabItems() as { tabText }, index}
         <VtmnTabsItem
           on:click={() => action('click')(index.toString())}
-          selected={index === 0 ? true : false}
+          aria-selected={index === 0 ? true : false}
```
thibault-mahe, line 29: `index === 0` is already a boolean expression. Maybe it could be simpler:

```
aria-selected={index === 0}
```

packages/sources/css/src/components/navigation/tabs/src/index.css
thibault-mahe: Indeed I already mention similar issues on this style. More globally, the CSS should not make assumptions on the HTML structure. But it's a minor change for now, I think the whole style would be reworked for the v1 :)

thibault-mahe (changes requested):

thibault-mahe:
@GaspardMathon @lauthieb @7studio @romuleald
I converted this PR to draft since there are still some changes to make
Feel free to reopen when the PR is ready for review again.
Thanks! :)

packages/showcases/css/stories/components/navigation/tabs/examples/medium.html
lauthieb: Agree!

packages/showcases/css/stories/components/navigation/tabs/examples/medium.html
romuleald: Yes, the is odd as the `aria-labelledby` id shouldn't be the same element

thibault-mahe:
I made some fixes for the CSS demo and examples since we need it now. if it is ok with you I'll merge as is and then reopen the issue for the Svelte / React / Vue fixes

thibault-mahe (approved):

thibault-mahe:
The rest of these developments, specific for framework integration, can be found here https://github.com/Decathlon/vitamin-web/pull/1256

PR #1238, opened 2022-08-19
fix(@vtmn/css, @vtmn/react, @vtmn/vue, @vtmn/svelte): linear progressbar animation

### LINKED ISSUE #1234: refactor(@vtmn/css, @vtmn/react, @vtmn/svelte, @vtmn/vue): the transition of `VtmnProgressbar` is not smooth
**Is your request related to a problem? Please describe.**

The `VtmnProgressbar` transition is not smooth when we are in variant "linear" & "determinate"

**Describe the solution you'd like**

https://user-images.githubusercontent.com/9600228/184312709-a1b9277e-5693-4969-96cf-2e65993fe8c7.mp4

**Additional context**

I think this is because we don't use `strokDashoffset` for the linear progressbar.

thibault-mahe:
## Changes description

- Rework linear progressbar with an animated progress
- "Boyscout" refacto on the progressbar: isolating aria-props, caching redundant variables & booleans, ...

## Context
- The linear progress bar didn't have an animated progress like the circular one has.
- Close #1234 

## Checklist

- No


thibault-mahe:
Regarding last commit on "harmonize showcases", here some context : 
- The SVG takes the size of its given `width` and `height` properties (if provided), or the size of its most direct parent with these properties;
- if no size is provided, the SVG can default to a 300 x 150 size . The consensus is not applied by all browsers, some (Firefox) let the svg taking up all the available space (see https://svgwg.org/specs/integration/#svg-css-sizing).
- This last point means that the progressbar showcases are different from one browser to another. To harmonize them I inserted the showcased component in a `div` with a `width` of 300px as it was already done for the Vue component.

packages/showcases/css/stories/components/indicators/progressbar/examples/overview.html
```diff
@@ -1,51 +1,54 @@
...
+        <span aria-live="assertive">60%</span>
+      </div>
+      <svg>
+        <line
+          class="vtmn-progressbar_indicator"
+          x1="0"
+          x2="100%"
+          y1="50%"
+          y2="50%"
+          style="transform: translateX(-50%)"
```
lauthieb, line 22:
```suggestion
          style="transform: translateX(-40%)"
```

packages/sources/vue/src/components/indicators/VtmnProgressbar/VtmnProgressbar.vue
```diff
@@ -69,9 +81,9 @@ export default /*#__PURE__*/ defineComponent({
   <div
     :class="classes"
     role="progressbar"
-    :aria-valuemin="0"
-    :aria-valuemax="100"
-    :aria-valuenow="value"
+    :aria-valuemin="status === 'determinate' ? 0 : undefined"
+    :aria-valuemax="status === 'determinate' ? 100 : undefined"
+    :aria-valuenow="status === 'determinate' ? value : undefined"
```
lauthieb, line 86: You can add these expressions in the return of the `setup` function.

packages/sources/vue/src/components/indicators/VtmnProgressbar/VtmnProgressbar.vue
```diff
@@ -69,9 +81,9 @@ export default /*#__PURE__*/ defineComponent({
...
     :aria-labelledby="labelId ? labelId : undefined"
```
lauthieb, line 87: For spreading `ariaProps` you can investigate: https://github.com/vuejs/core/issues/4105

lauthieb (changes requested):
Thanks a lot for all these changes @thibault-mahe, just few suggestions :)

packages/sources/svelte/src/components/indicators/VtmnProgressbar/VtmnProgressbarLinear.svelte
```diff
@@ -69,9 +72,10 @@
     <line
       class="vtmn-progressbar_indicator"
       x1="0"
-      x2={`${indeterminate ? 100 : progress}%`}
+      x2="100%"
       y1="50%"
       y2="50%"
+      style:transform={indeterminate ? null : `translateX(${progress - 100}%)`}
```
Tlahey, line 78: Not fan to et directly the style on the node.
I think it's better to set a var `--vtmnProgressbar-progress` (or other) and apply it on a style 

```css
.vtmn-progressbar_indicator {
  transform: translateX(var(--my_var, '0%'))
}
```

packages/sources/svelte/src/components/indicators/VtmnProgressbar/VtmnProgressbarLinear.svelte
```diff
@@ -45,17 +45,20 @@
     indeterminate && `vtmn-progressbar--indeterminate`,
     className,
   );
+
+  $: ariaProps = {
+    'aria-labelledby': labelId ? labelId : undefined,
```
Tlahey, line 50: @lauthieb I don't know if a rework is needed here, but finally, we can use directly the `aria-labelledby` from the `$$restProps` ? 

Or do you think it's better to keep like that ? 🤔 

Otherwise, we can set the `export let labelId = undefined` and remove the ternary line 50 :)

packages/sources/svelte/src/components/indicators/VtmnProgressbar/VtmnProgressbarLinear.svelte
thibault-mahe: Hi Antoine, thanks for the review :) 
Can you elaborate more on this please ?
I do understand that inline styling is not the best pattern in most cases, but I think that in this specific case it is very common:
- the `progress` value is a reactive prop of the component: to pass it to the CSS we will still have to inline the custom property or go through a heavier `setProperty` in JS
- Most of the progress bar uses this approach, including some famous design system. (But I understand that this kind of argument is not always a good argument ;) )

I would be happy to hear more from you on this

packages/sources/svelte/src/components/indicators/VtmnProgressbar/VtmnProgressbarLinear.svelte
thibault-mahe: Hi @Tlahey,
the idea behind this rework was to harmonize code with Vue and React. 
I agree with you to explicitly default the labelId to `undefined` directly in the props though. A default value is set on the parent `VtmnProgressbar` element, but not explicit in this `VtmnProgressbarLinear`.

packages/sources/svelte/src/components/indicators/VtmnProgressbar/VtmnProgressbarLinear.svelte
Tlahey: Of course ! 

In fact, we got some issue with a style than the ponderation of the css property override the customClass passed on the component.

It's also a good practice in order to keep all style content on the same scope (css files).

On you svelte file you can have 

```svelte
<script>
   // [...]
</script>

<svelte:style>
    .progress_indicator {
        transform: translateX(var(--my_var, '0%'));
    }
</svelte:style>

<line
      class="vtmn-progressbar_indicator progress_indicator"
      style:transform={indeterminate ? null : `translateX(${progress - 100}%)`}
/>
```

On the second hand, have an inline style are bad for accessibility : 
https://www.nomensa.com/blog/css-inline-styles-and-why-they-are-considered-harmful-accessibility
But of course, for a transform, I think it's questionable :) 

In conclusion (for my opinion), use var instead of direct inline style : 
- Improve the ponderation of the style
- Easily overridable with className
- All the style are scope on the css file or `<svelte:style>` node

thibault-mahe:
Thanks for your reviews @lauthieb @Tlahey, I made some changes accordingly:

-  progressbar's animation is now made with a custom property set on the HTML/JSX, and the variable is then directly used in the css of the component (in `packages/sources/css/...progressbar/...`)
- the `labelId` has now a default `undefined` value to avoid using a ternary when its value is used
- progressbar state boolean and ariaProps are now put in cache inside the setup constructor in Vue component (similar to the React one)
- In Svelte, I added the class `vtmn-progressbar_variant--linear` on linear component. It was missing (the class is set on both React and Vue components) and is now needed for the CSS. It might be a breaking change 😬 Can you check on this and tell me @Tlahey please ?

Thanks guys!

lauthieb (dismissed):
Great job @thibault-mahe :tada:

lauthieb:
Hi @Tlahey, can you please check again? Thanks in advance :)

packages/sources/css/src/components/indicators/progressbar/src/index.css
```diff
@@ -41,6 +41,13 @@
   align-items: center;
 }
 
+/* Linear */
+.vtmn-progressbar_variant--linear .vtmn-progressbar_indicator {
+  --vtmn-progressbar_progress-transform: unset;
+
+  transform: var(--vtmn-progressbar_progress-transform);
```
Tlahey, line 48: What do you think about used the second property from `var` to set a default value ? 
https://developer.mozilla.org/fr/docs/Web/CSS/var

packages/sources/react/src/components/indicators/VtmnProgressbar/VtmnProgressbar.tsx
```diff
@@ -149,11 +149,11 @@ export const VtmnProgressbar = ({
...
               style={
-                isDeterminate
-                  ? {
-                      transform: `translateX(${progress - 100}%)`,
-                    }
-                  : {}
+                {
+                  '--vtmn-progressbar_progress-transform': isDeterminate
+                    ? `translateX(${progress - 100}%)`
+                    : 'unset',
```
Tlahey, line 155: What do you think about this short solution ?

```html
style={isDeterminate && `translateX(${progress - 100}%)`}
```

packages/sources/svelte/src/components/indicators/VtmnProgressbar/VtmnProgressbarLinear.svelte
```diff
@@ -75,7 +76,8 @@
       x2="100%"
       y1="50%"
       y2="50%"
-      style:transform={indeterminate ? null : `translateX(${progress - 100}%)`}
+      style={'--vtmn-progressbar_progress-transform:' +
+        (indeterminate ? 'unset' : `translateX(${progress - 100}%)`)}
```
Tlahey, line 80:
```html 
style={!indeterminate && `translateX(${progress - 100}%)`}
```

packages/sources/vue/src/components/indicators/VtmnProgressbar/VtmnProgressbar.vue
```diff
@@ -55,96 +55,109 @@ export default /*#__PURE__*/ defineComponent({
...
         y2="50%"
-        :style="
-          status === 'determinate'
-            ? { transform: `translateX(${progress - 100}%)` }
-            : {}
-        "
+        :style="{
+          '--vtmn-progressbar_progress-transform': isDeterminate
+            ? `translateX(${progress - 100}%)`
+            : 'unset',
```
Tlahey, line 131: Same :)

Tlahey:
@thibault-mahe thanks for your enhancement ! 🤩 
I suggest you littles changes on your PR, I'm available to discuss :)

packages/showcases/css/stories/components/indicators/progressbar/examples/linear.html
```diff
@@ -207,13 +213,19 @@
...
+      );
     }
 
     function resetProgress(container, progress, label) {
       label.textContent = '0%';
       container.setAttribute('aria-valuenow', '0');
-      progress.style.transform = `translateX(-105%)`;
+      progress.style.setProperty(
+        '--vtmn-progressbar_progress-transform',
+        `translateX(-100%)`,
```
GaspardMathon, line 227:
```suggestion
        `translateX(-105%)`,
```

The 5% added make the linear progressbar really empty at 0%.

packages/showcases/css/stories/components/indicators/progressbar/examples/overview.html
```diff
@@ -1,51 +1,54 @@
...
-    aria-valuemax="100"
-    aria-valuenow="60"
-  >
-    <div class="vtmn-progressbar_label">
-      <span id="progress-bar-overview-1">Loading</span>
-      <span aria-live="assertive">60%</span>
+<div style="width: 300px">
+  <div class="block">
+    <div
+      class="vtmn-progressbar_container vtmn-progressbar_variant--linear"
```
GaspardMathon, line 4: In the overview examples I used to be as compact as possible wit h the components. That's why sometimes you don't have the vtmn-progressbar_variant-- or vtmn-progressbar_size-- classes to display the default css styles. So for this example I would remove it too.
```suggestion
      class="vtmn-progressbar_container"
```

packages/sources/react/src/components/indicators/VtmnProgressbar/VtmnProgressbar.tsx
```diff
@@ -74,109 +80,128 @@ export const VtmnProgressbar = ({
...
+            <span className="vtmn-progressbar_label" aria-live="assertive">
+              {progress}%
+            </span>
+          )}
+
+          {imageSrc !== undefined && (
+            <img
+              className="vtmn-progressbar_image"
+              src={imageSrc}
+              alt={imageAlt ?? ''}
```
GaspardMathon, line 179: nice :)

GaspardMathon (changes requested):

packages/showcases/css/stories/components/indicators/progressbar/examples/linear.html
thibault-mahe: But "5%" doesn't really make sense and is a [magic number](https://css-tricks.com/magic-numbers-in-css/), it will add more confusion I think. Why 105 ? Why not 103 or 110 ?

Various alternatives are possible here, but first: why is the progressbar not empty on 0%? Because the `stroke-with` property of our line makes it a little bigger on each side. My guess is that if the `stroke-with` is set to X px, then it adds x/2 on both sides. We could make it explicit:
```
`translateX(calc(-0.5rem / 2 - 100%))`

// or even
`translateX(calc(-var(--vtmn-progressbar_progress-stroke-width) / 2 - 100%))`
```

Another possibility could be to set `stroke-with` as a variable, for instance : 
```
progress.style.setProperty(
        '--vtmn-progressbar_progress-stroke-width',
        progress === 0 ? "0" : "0.5rem"
)
```

Another possibility could be to ignore this thin line at the beginning when the progress is 0. But then it should be validated by designers I guess.

What do you think @GaspardMathon ?

packages/sources/css/src/components/indicators/progressbar/src/index.css
thibault-mahe: I think it's a good idea, I'll make the fix., thank you!
Just in case you didn't know, `var` already defaults to `unset` if the custom property doesn't exist. But in this specific case I think it is better to explicit this.

packages/sources/react/src/components/indicators/VtmnProgressbar/VtmnProgressbar.tsx
thibault-mahe: Do you mean
```
style={isDeterminate && {
    '--vtmn-progressbar_progress-transform': `translateX(${progress - 100}%)`
}}

```
?

I agree that this way the default value is managed by the css directly

romuleald: Why do we keep SVG for the linear progress? Seems to me to be too much complexity instead of using a `div` tag and a hidden `progress` for a11y.

thibault-mahe: Hi @romuleald, thanks for your comment!
SVG doesn't bring to much complexity here, after all it is juste an `svg` tag with a `line` tag, not that different from having a `div` inside a `div`. At the end of the day I agree that simpler shapes should be drawn with CSS instead of SVG though. In both cases the line/div is animated with a `transform` property cached in a CSS variable, complexity remains the same.

Finally, note that SVG is kept here to keep the consistency between linear progress and circular progress (this one is easier to draw with SVG) and, above all, the purpose of this issue is only to fix a minor change. I agree with the `progress` pattern, probably the direction we'll take when the component is reworked in the v1.

thibault-mahe:
Last commit fixes:
- removing "unset" value set on component. The default value is then directly managed by the CSS. cc @Tlahey 
- visually hide the indicator stroke when the progress is null. This one is manage with a scale value set to 0, it is the simpler and more elegant solution I came with. What do you think @GaspardMathon ?

 Hi @romuleald, feel free to review this changes as well if you want to :)

packages/showcases/css/stories/components/indicators/progressbar/examples/linear.html
GaspardMathon: I am more for the first option as it could be implemented in CSS easier.

GaspardMathon: So at this time, we can either let the property set to 100% or we can reopen an issue to fix the design if it is a problem, what do you think @thibault-mahe ?

thibault-mahe: Did you see my last message and commit ? I managed the issue with (what I think is) a more elegant fix. Thanks to the scale value of a transform, I can show / hide the visible progress. I managed to put the scale to 0 when progress is 0, or to 1 when progress is > 0.

You can see it here for instance: https://github.com/Decathlon/vitamin-web/blob/fix/linear-progressbar-animation/packages/showcases/css/stories/components/indicators/progressbar/examples/linear.html#L226

GaspardMathon: Outdated

GaspardMathon (approved):

lauthieb (approved):

lauthieb (approved):

PR #1243, opened 2022-08-24
feat(@vtmn/svelte): add `on:input` and `on:change` on `VtmnTextInput` component

ChrisHendrX:
## Changes description
Added on:input and on:change directives

## Context
With the previous implementation, the VtmnTextInput couln't handle a specific action whenever the value changes (on:input) like the native HTML input does.
For example, in my case I couldn't perform validations on the fly whenever the textfield changes

## Checklist
- Add on:input and on:change on theTextInput implementation to inherit the native HTML input behaviour
- Updated the TextInput's Component Story Format to handle these actions
- Added unit test to check if the on:input event is called once on every input change

No

### DIFF
--- a/packages/sources/svelte/src/components/forms/VtmnTextInput/VtmnTextInput.svelte
+++ b/packages/sources/svelte/src/components/forms/VtmnTextInput/VtmnTextInput.svelte
@@ -84,6 +84,8 @@
 {#if multiline}
   <textarea
     bind:value
+    on:input
+    on:change
     class={componentClass}
     {id}
     {disabled}
@@ -94,6 +96,8 @@
   <div class="vtmn-text-input_container">
     <input
       bind:value
+      on:input
+      on:change
       class={componentClass}
       type="text"
       {id}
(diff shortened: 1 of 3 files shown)


packages/sources/svelte/src/components/forms/VtmnTextInput/VtmnTextInput.svelte
```diff
@@ -94,6 +96,8 @@
   <div class="vtmn-text-input_container">
     <input
       bind:value
+      on:input
```
lauthieb, line 99: Hi @ChrisHendrX thanks for your proposition.
In Svelte, it's not possible to spread all events? Instead of having to think about all the events that we need to forward... In React, we can thanks to {...props}
Thanks in advance.

thibault-mahe (approved):

GaspardMathon:
@lauthieb need a manual approval for first PR it seems.

packages/sources/svelte/src/components/forms/VtmnTextInput/VtmnTextInput.svelte
lauthieb: @ChrisHendrX, what's your opinion please regarding my comment just above? Thanks.

lauthieb (approved):
That's ok for me, thanks Christophe!

thibault-mahe (approved):

Tlahey (approved):

PR #1245, opened 2022-08-25
feat(@vtmn/css, @vtmn/svelte, @vtmn/react, @vtmn/vue): add rel attribute for blank link targets to prevent security issues

### LINKED ISSUE #1242: feat: security enhancement on `VtmnLink` component
Hello,

I've found a potential security issue with the `VtmnLink` component.
See this post -> https://twitter.com/brianthiely/status/1562102291502665728?s=21&t=27WqGeqocjEcRezD05-ziA

We need to add a `noopener noreferrer` when target is `_blank`

GaspardMathon:
## Changes description

- Add `rel="noopener noreferrer"` to every link with an explicit `target` value set to `_blank` to prevent security issues

## Context

- Close #1242 

## Checklist

- No

## Other information

Thanks @Tlahey for the issue 🔥


lauthieb (changes requested):
Hi @GaspardMathon,

Thank you very much for this PR that corrects this security problem in our showcases. However, I don't think it completely fixes the issue raised by @Tlahey.

Indeed, we would also need to handle this automatically within our `VtmnLink` components on our different libraries `@vtmn/react`, `@vtmn/svelte` `& @vtmn/vue`. 
If `target='_blank``, then force the addition of `rel="noopener noreferrer"` inside its markup.

What do you think @GaspardMathon & @Tlahey?

Thanks :)

Tlahey:
Of course @lauthieb !
This security issue must be include inside the components :) 

if target = _blank and rel doesn't include noopener add noopener / rel doesn't include noreferrer add noreferrer

thibault-mahe (changes requested):
Agreed with @lauthieb & @Tlahey on the `VtmnLink` needed change.
Thanks!

Tlahey:
@GaspardMathon I will provide the changes on the svelte part :)

packages/sources/react/src/components/actions/VtmnLink/VtmnLink.tsx
```diff
@@ -17,6 +17,21 @@ export interface VtmnLinkProps extends React.ComponentPropsWithoutRef<'a'> {
...
+   */
+  target: string;
+
+  /**
+   * The rel attribute of the link.
+   * @type {string}
+   * @defaultValue undefined
+   */
+  rel?: string;
+
```
thibault-mahe, line 34: I'm not sure that we should rewrite the whole (or part of the) HTML link standard API here:
- User should be able to use any legal HTML link attribute on our `VtmnLink`. User should just write attributes and this attributes should be on the rendered link ;
- Optional HTML link attribute should not be made mandatory for no reason in our `VtmnLink` (with maybe some accessibility exception).

Meaning that `target` and `rel`, that are already part of the link api, should not be a component props.
Instead we should maybe just "intercept" some native attribute to edit their values, when needed.

For instance :
```
export const VtmnLink = ({
   // no "rel" nor "target" props here 
  ...,
  // if user add native rel or target attributes, they will be stored in the following:
  ...props
}: VtmnLinkProps) => {

  // "Intercept" attributes to edit it
  const relAttrValues = Array.from(
    new Set([
      // If component has a given "rel" attribute, keep values
      ...(props?.rel?.split(' ') ?? []),
      // If comp has "target" set to "blank", add some values
      ...(props?.target === '_blank' ? ['noopener', 'noreferrer'] : []),
    ]),
  ).join(' ').trim();

  return (
    <a
      ...
      {...props}
      {...(relAttrValues ? { rel: relAttrValues } : {})}
    >
      ...
    </a>
  );
};
```

thibault-mahe (changes requested):

thibault-mahe (approved):

thibault-mahe (approved):

lauthieb (approved):

lauthieb (approved):

lauthieb:
@thibault-mahe for me that's ok.
If we are sure there's no breaking change, you can merge it directly. If there are breaking changes, please merge & inform all our consumers.
Thanks!

PR #1258, opened 2022-09-16
fix(@vtmn/svelte): class on the `VtmnSearch`

Tlahey:
Fix an issue on the `VtmnSearch`.

Before : 
- `className` export should be set to pass class to the `VtmnSearch` component

Now : 
- `class` inject the component's class to the `VtmnSearch` component


thibault-mahe (approved):
Seems legit and more "standard/future-proof" I agree.
Thanks for you contribution @Tlahey :)

PR #1272, opened 2022-10-10
chore: upgrade dependencies

lauthieb:
## Checklist

- No


thibault-mahe:
Seems ok to me, but one thing we haven't discussed yet is the range versioning. Here we are using the default caret (^) to allow npm to widen the range of accepted versions to the minor or patch ones. It means 3 things i think:
- 🙈 we don't have full control on the dependencies versions
- 🙀 our build are not reproductible since they depend on the time we run the npm install command (it could be ugly if we want to cleanly roll back)
- 👍 contributors don't have to wait for us updating a breaking version to use the patch of a dependency

Are we ok with that ? We could also go with exact versioning as an alternative with other drawbacks I guess

lauthieb:
@thibault-mahe for me that's ok but, what's your opinion @thibault-mahe?

thibault-mahe:
I would say that for a widely distributed and used library like Vitamin it would be better to fix the dependencies. Some libraries (possibly among our dependencies) sometimes break stuff on minor versions making it impossible for our consumers to use our components. By fixing the dependencies we keep full responsibility for the proper functioning of our library. But there is drawbacks too.

Let's keep the current way for now, i'll ask our SIG and frontend community about the subject :)

PR #1292, opened 2022-10-25
refactor(@vtmn/icons, @vtmn/assets): `svgo` config to add `removeDimensions` plugin

lauthieb:
## Changes description

I propose to add the [`removeDimensions`](https://github.com/svg/svgo/blob/main/plugins/removeDimensions.js) plugin from [`svgo`](https://github.com/svg/svgo) in order to have a better visualisation of our different icons & assets.

## Before

CleanShot 2022-10-25 at 08 40 24@2x

CleanShot 2022-10-25 at 08 41 11@2x

## After

CleanShot 2022-10-25 at 08 40 44@2x

CleanShot 2022-10-25 at 08 41 42@2x

## Checklist

- No

I've checked by launching our different showcases and tested the icon font with some icons like "Button" with left/right & only icon. But, please if you can ensure this, that could be great. Thanks 🙏


thibault-mahe (approved):

lauthieb:
@thibault-mahe as we didn't really use these heights and widths because we permit to use them in 16x16, 32x32 or 64x64, I think it is more related to `viewBox` and let our consumers choose the height/width with these specific attributes instead of choosing one of these by default.
For performance issues, I don't really know what this change will cause.

PR #1304, opened 2022-11-10
feat(@vtmn/svelte, @vtmn/react, @vtmn/vue, @vtmn/css): add heading level on title + title wraps correctly

### LINKED ISSUE #1305: bug: cards a11y & visual issues
**Describe the bug**

The VtmnCards currently in production (on Svelte) have some visual (the heading is not wrapping) and accessibility (headings are note semantic) issues.

image

**Steps to reproduce**
The issues can be seen directly on decathlon.fr or on the [CSS showcase](https://decathlon.github.io/vitamin-web/@vtmn/showcase-css/?path=/story/components-structure-card--overview) or in the [Svelte showcase](https://decathlon.github.io/vitamin-web/@vtmn/showcase-svelte/?path=/story/components-structure-vtmncard--overview)

**Expected behavior**

- allow component consumer to choose the heading level of the title so it can fit hierarchically in the page flow
- remove the `min-inline-size` of the title which is forcing the size of the card and preventing the text to wrap

**Browsers affected**

all

**Version affected**
@vtmn/css: 0.85.17 and current front packages

Tlahey:
image

Add new input `titleElement` in order to switch to an heading level (for accessibility and SEO).

In order to not have regression, we have to keep a span on the default element.

- Resolve #1305


Tlahey:
As see on the best practices -> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#best_practices

"The best way to use this role is to not use it at all". So I think we have to possibility to switch the element on svelte easily.

lauthieb:
Hi @Tlahey, thanks for this PR.
Do you think it could be solved at CSS stage instead of Svelte?

GaspardMathon:
@lauthieb For the fix of the text wrapping it can be resolved by CSS. For the accessibility of the title element, it is not a breaking change, it is a new feature.

packages/showcases/svelte/stories/components/structure/VtmnCard/VtmnCard.stories.svelte
```diff
@@ -13,7 +13,19 @@
   title="Components / Structure / VtmnCard"
   component={VtmnCard}
   {parameters}
-  {argTypes}
+  argTypes={{
+    ...argTypes,
+    headingLevel: {
+      type: { name: 'string', required: false },
```
Tlahey, line 19:
```suggestion
      type: { name: 'number', required: false },
```

packages/showcases/svelte/stories/components/structure/VtmnCard/VtmnCard.stories.svelte
```diff
@@ -13,7 +13,19 @@
...
+      description: 'Heading level of the card title',
+      defaultValue: undefined,
+      control: {
+        type: 'select',
+        options: [undefined, 1, 2, 3, 4, 5, 6],
+        default: undefined,
```
thibault-mahe, line 25: do we want an undefined heading level ? 2 or 3 could be a valid default value don't you think ?

packages/sources/svelte/src/components/structure/VtmnCard/VtmnCard.svelte
```diff
@@ -59,7 +65,13 @@
...
   <div class={componentContentClass}>
     {#if title}
-      <span class="vtmn-card_content--title">{title}</span>
+      <span
+        role={headingLevel && 'heading'}
+        aria-level={headingLevel && headingLevel}
+        class="vtmn-card_content--title"
+      >
+        {title}
+      </span>
```
thibault-mahe, line 74: A possible semantic alternative: 
```
const Heading = `h${headingLevel}`;
```
and then
```
<Heading class="vtmn-card_content--title">
       {title}
</Heading>
```

packages/showcases/svelte/stories/components/structure/VtmnCard/VtmnCard.stories.svelte
GaspardMathon: We have to keep a value undefined not to break the current card that have a 'span'.

packages/sources/svelte/src/components/structure/VtmnCard/VtmnCard.svelte
GaspardMathon: Good idea

packages/showcases/svelte/stories/components/structure/VtmnCard/VtmnCard.stories.svelte
thibault-mahe: I don't understand. If there is a default value then the prop is optional then existing card without the prop and with the span would not break, no ? Am I missing something ?

packages/showcases/svelte/stories/components/structure/VtmnCard/VtmnCard.stories.svelte
GaspardMathon: thanks

packages/showcases/svelte/stories/components/structure/VtmnCard/VtmnCard.stories.svelte
GaspardMathon: Yes, you are right. It shouldn't impact current Cards

packages/sources/svelte/src/components/structure/VtmnCard/VtmnCard.svelte
Tlahey: Not possible, to do that it's my first commit with the `<svelte:element>`

Tlahey:
@GaspardMathon set heading level to 2 should introduce side effects on the code. 
We want to keep the span by default on svelte and offer the possibility to set title as h1 / h2 ...

packages/sources/svelte/src/components/structure/VtmnCard/VtmnCard.svelte
```diff
@@ -24,12 +24,20 @@
...
+   */
+  export let headingLevel = 2;
+
   let className = undefined;
   /**
    * @type {string} Custom classes to apply to the component.
    */
   export { className as class };
 
+  const Heading = `h${headingLevel}`;
```
Tlahey, line 39: We need to make a reactive var here like 

```suggestion
  $: titleElement = headingLevel > 0 && headingLevel <= 6 && `h${headingLevel}` || 'span';
```

packages/sources/svelte/src/components/structure/VtmnCard/VtmnCard.svelte
```diff
@@ -24,12 +24,20 @@
    */
   export let contentOpaque = false;
 
+  /**
+   * @type {1|2|3|4|5|6} Heading level of card title
+   * @default 2
+   */
+  export let headingLevel = 2;
```
Tlahey, line 31: We have to keep the default behaviours. We already used the VtmnCard on the ecomm website.

```suggestion
  /**
   * @type {1|2|3|4|5|6} Heading level of card title
   */
  export let headingLevel;
```

packages/sources/svelte/src/components/structure/VtmnCard/VtmnCard.svelte
```diff
@@ -59,7 +67,9 @@
   {/if}
   <div class={componentContentClass}>
     {#if title}
-      <span class="vtmn-card_content--title">{title}</span>
+      <svelte:element this={Heading} class="vtmn-card_content--title">
```
Tlahey, line 70:
```suggestion
      <svelte:element this={titleElement} class="vtmn-card_content--title">
```

packages/showcases/core/csf/components/structure/card.csf.js
```diff
@@ -42,4 +42,14 @@ export const argTypes = {
       type: 'text',
     },
   },
+  headingLevel: {
+    type: { name: 'number', required: false },
+    description: 'Heading level of the card title',
+    defaultValue: 2,
```
Tlahey, line 48: Should be different on svelte

Tlahey, line 48: Not a default value for svelte

packages/showcases/svelte/stories/components/structure/VtmnCard/VtmnCard.stories.svelte
```diff
@@ -13,7 +13,9 @@
   title="Components / Structure / VtmnCard"
   component={VtmnCard}
   {parameters}
-  {argTypes}
+  argTypes={{
+    ...argTypes,
+  }}
```
Tlahey, line 18:
```suggestion
  {argTypes}
```

thibault-mahe:
As we discussed there is 2 options here : 

- forcing the use of heading with a default heading to 2. The only side effect this could have on former component version is visual and could therefore be anticipated. With the specific title class we have it should be easy to set a fixed font-weight, font-size, margin, etc.
- allowing the use of a span for former component. It is not a11y compliant but 1) we make sure to not cause any breaking change or visual regression & 2) the the component will be reworked anyway with the v1 of Vitamin.

I agree with both possibilities, I prefer the first one, decision is yours ;)

lauthieb:
Hi everyone,
Thanks for this. I agree with @thibault-mahe & preference for solution 1 mentioned [here](https://github.com/Decathlon/vitamin-web/pull/1304#issuecomment-1310485723).
As I can see, it's a global PR for Vue, React, Svelte & CSS. Please, we must not forget any community bookstore and let's avoid the breaking changes.
Thank you :)
Laurent

thibault-mahe:
Hi everyone, there's been no updates on this PR for 6 days.
Since we should exercising extra caution on this subject, let's follow @Tlahey recommendations and allowing the use of span on the svelte version.
@GaspardMathon will you have some times to answering the missing requirements or do you want me to take care of it ?
Thanks :)

GaspardMathon:
@thibault-mahe @lauthieb @Tlahey I am on it. Sorry for the delay, I only work on vitamin subjects on Thurdays :)

packages/showcases/core/csf/components/structure/card.csf.js
GaspardMathon: done

GaspardMathon:
Changes made to be sure: all libraries but Svelte have a default h2 titleElement, titleElement that can be chosen between all the HTML title tags possible. Svelte just has a default span tag moreover.

packages/sources/svelte/src/components/structure/VtmnCard/VtmnCard.svelte
```diff
@@ -24,12 +24,24 @@
...
   export let contentOpaque = false;
 
+  /**
+   * @type {1|2|3|4|5|6} Heading level of card title
+   * @default 2
+   */
+  /**
+   * @type {1|2|3|4|5|6} Heading level of card title
+   */
+  export let headingLevel;
```
Tlahey, line 34: defined 2 times

packages/showcases/core/csf/components/structure/card.csf.js
```diff
@@ -42,4 +42,13 @@ export const argTypes = {
...
   },
+  headingLevel: {
+    type: { name: 'number', required: false },
+    description: 'Heading level of the card title',
+    control: {
+      type: 'range',
+      min: 1,
+      max: 6,
+    },
+  },
```
Tlahey, line 53: You have to set the headingLevel required for react / vue but not required on svelte

Tlahey, line 53: Please set default value = 2 but override as span only for svelte

packages/sources/react/src/components/structure/VtmnCard/VtmnCard.tsx
```diff
@@ -47,9 +53,12 @@ export const VtmnCard = ({
   img,
   fullImage = false,
   contentOpaque = false,
+  headingLevel = 2,
   children,
   className,
 }: VtmnCardProps) => {
+  const Heading = `h${headingLevel}` as keyof JSX.IntrinsicElements;
```
Tlahey, line 60: We need to can you check if the headingLevel are correctly set >0 <=6 ?

packages/sources/svelte/src/components/structure/VtmnCard/test/VtmnCard.spec.js
```diff
@@ -56,6 +56,13 @@ describe('VtmnCard', () => {
...
       expect(getImg(container)).toBeVisible();
     });
+    test('Should set the title as h1', () => {
+      const { getByRole } = render(VtmnCard, {
+        title: 'unit test',
+        headingLevel: 1,
+      });
+      expect(getByRole('heading', { level: 1 })).toBeVisible();
+    });
   });
```
Tlahey, line 66: Can you please add a test by default the title are into a span ?
Thanks :)

GaspardMathon: All changes mentioned have been made. Thanks @Tlahey ;)

lauthieb (approved):

thibault-mahe (approved):

Issue #1313, opened 2022-11-24
bug(@vtmn/assets): wrong flag display on Firefox/Chrome


Jerome1337:
**Describe the bug**

After a fresh upgrade of `@vtmn/assets` packages some flags are not correctly render on Firefox/Chrome. Flags not correctly renders are differents depending of the browser.

For example on Firefox:
<img width="142" alt="Screenshot 2022-11-24 at 2 15 02 PM" src="https://user-images.githubusercontent.com/7968649/203793816-dccd0b40-9318-4f4a-8c16-17c7a7f58354.png">
<img width="163" alt="Screenshot 2022-11-24 at 2 14 52 PM" src="https://user-images.githubusercontent.com/7968649/203793814-5ac107c9-e89c-4f4f-891c-35fa4bd5f156.png">

And Chrome:

<img width="180" alt="Screenshot 2022-11-24 at 2 15 11 PM" src="https://user-images.githubusercontent.com/7968649/203793948-0ef15661-0166-40c7-ab7e-44407a3b6d9b.png">
<img width="148" alt="Screenshot 2022-11-24 at 2 15 16 PM" src="https://user-images.githubusercontent.com/7968649/203793949-9a304a8c-f984-4862-a8c7-cbd941966899.png">

**Steps to reproduce**

Use the flag sprite and display a Poland flag on Firefox or Colombia flag on Chrome.

**Expected behavior**

Flags should be corrects on every browser.

**Browsers affected**

- Chrome
- Firefox
- Safari

**Version affected**

`@vtmn/assets: 0.3.0`

thibault-mahe:
Hi @Jerome1337 , thanks for reaching us :) 
Could you be more specific on the way your are consuming the flags please ? Are you using an `img` tag ? the svg `use` ? etc. Maybe if you can provide us with a demo it could be easier to debug :) 

I try to use both the Poland and Colombia flag on my side (@vtmn/assets: 0.3.0 with NextJs) and it seems to work well on both Chrome and Firefox.

Thanks!

Jerome1337:
Hi @thibault-mahe,

We use the Sprite like this

```js
import flags from '@vtmn/assets/dist/assets/sprite/assets.svg';

...

// item.label is for example France...
// item.value is for example fr or de...
<svg width="22" height="19" fill="#001018" key={item.value} id={`ChooseCountry-${item.label.toUpperCase()}`}>
  <use href={`${flags}#flag-${item.value}`} />
</svg>
[...]

PR #1315, opened 2022-11-28
fix: rework assets sprite building

thibault-mahe:
## Changes description

Remove all useless attributes and tags in svg sprite causing bugs in some browsers

## Context

When using the the svg <use> element from an external reference (in our case the sprite/assets.svg#some-flag), the original svg is cloned inside a shadow root to keep import in isolation. It seems that sometimes Chrome doesn't know how to handle a local fetch with url() inside a remote file and do not honor the original defs element.

This PR suggests to remove all url() since they seems useless and only bring bugs...

See more comment on the affiliated issue: https://github.com/Decathlon/vitamin-web/issues/1313

Also, the way browsers are differently handling `<defs> `content in external svgs & omitting many definitions like clip paths, gradients, masks or styles, ... is a known bug, at least since 2012. See for instance:
- Stack overflow issue: 
     - https://stackoverflow.com/questions/74128326/svg-with-defs-from-external-symbol-sheet-not-loading-in-firefox-or-safari/
     - https://stackoverflow.com/questions/74292618/svg-use-ignores-gradient-styling
- Chromium bug trackers : [issue 1385248](https://bugs.chromium.org/p/chromium/issues/detail?id=1385248&q=svg%20use&can=2), [issue 1094613](https://bugs.chromium.org/p/chromium/issues/detail?id=1094613), [issue 951265](https://bugs.chromium.org/p/chromium/issues/detail?id=951265), [issue 729599](https://bugs.chromium.org/p/chromium/issues/detail?id=729599), [issue 229081](https://bugs.chromium.org/p/chromium/issues/detail?id=229081), ...

## Checklist

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch**. Please, don't request directly from your main!
- [ ] Check commits & PR names matches our requested structure. It must follow the https://www.conventionalcommits.org pattern.
- [ ] Check your code additions will fail neither code linting checks.
- [ ] I have reviewed the submitted code.
- [ ] If it includes design changes, please ask for a review with a core team designer.

TO BE TESTED

- Yes
- No

### DIFF
--- a/packages/sources/assets/package.json
+++ b/packages/sources/assets/package.json
@@ -25,7 +25,7 @@
     "prebuild": "shx rm -rf dist",
     "build": "yarn build:svg && yarn build:json && yarn build:sprite",
     "build:json": "shx rm -rf ./dist/assets/json && shx mkdir ./dist/assets/json && fantasticon",
-    "build:sprite": "shx rm -rf ./dist/assets/sprite && shx mkdir ./dist/assets/sprite && spritesh -i dist/assets/svg -o dist/assets/sprite/assets.svg && node src/scripts/add-sprite-declaration.js",
+    "build:sprite": "shx rm -rf ./dist/assets/sprite && shx mkdir ./dist/assets/sprite && spritesh -i dist/assets/svg -o dist/assets/sprite/assets.svg && svgo dist/assets/sprite/assets.svg -o dist/assets/sprite/assets.svg --config ./svgo-sprite.config.js && node src/scripts/add-sprite-declaration.js",
     "build:svg": "node src/scripts/build-svg.js && svgo -f dist/assets/svg"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


thibault-mahe:
image

There is still an issue with some flags using a combination of `fill` and `url()`. And I can't just remove the fill attribute in this case since it is used with hard values most of the times.

i'll try with a regex to remove `fill=url(...`

thibault-mahe:
New update:  `fill=url(...` cannot be removed, too many flags are depending on it and linear gradient...
I'll look into alternative solutions...

thibault-mahe:
Update: still in draft, waiting for https://github.com/Decathlon/vitamin-design/issues/135

thibault-mahe:
Closing this, the icons will be reworked on the design part so we shouldn't have to rework them and our build process on our side.

Issue #1265, opened 2022-09-28
bug(@vtmn/css): `VtmnTabsItem` active border not showing when focusing outside component


MaelthiS:
**Describe the bug**

On the VtmnTabs component, there is a border that shows the active tab. When we focus outside the component, the border disapears.

**Steps to reproduce**

**Expected behavior**

The behavior should be : 
1. I click on the first tab of the VtmnTab
2. I click outside the component
3. The border to show the active tab should still be visible.

**Browsers affected**

**Version affected**

MaelthiS:
1. I go to https://decathlon.github.io/vitamin-web/@vtmn/showcase-react/?path=/docs/components-navigation-vtmntabs--overview
2. I click on the first tab of the VtmnTab
3. I click outside the component
4. The border to show the active tab should still be visible.

GaspardMathon:
Hello @MaelthiS, thanks for reaching us.

The `tabs` component has been reworked lately.
I tested your behaviour and the active tab remains active even if I click outside of the component.

Be aware that only the [demo showcase](https://decathlon.github.io/vitamin-web/@vtmn/showcase-css/?path=/story/components-navigation-tabs--demo) is accessible. The other ones are only visual or limited (because no javascript).

Can you confirm that this problem is resolved so that I can close this issue ?

lauthieb:
Hello,

I close this issue in favor to the answer of @GaspardMathon.

> Hello @MaelthiS, thanks for reaching us.
> 
> The `tabs` component has been reworked lately. I tested your behaviour and the active tab remains active even if I click outside of the component.
> 
> Be aware that only the [demo showcase](https://decathlon.github.io/vitamin-web/@vtmn/showcase-css/?path=/story/components-navigation-tabs--demo) is accessible. The other ones are only visual or limited (because no javascript).
> 
> Can you confirm that this problem is resolved so that I can close this issue ?

@MaelthiS feel free to reopen it if this is still not ok for you.

Have a nice day :-)

Laurent

thibault-mahe:
Hi @MaelthiS, all the demos in the CSS showcase are made with vanilla JS inserted inside a script tag: https://github.com/Decathlon/vitamin-web/blob/main/packages/showcases/css/stories/components/navigation/tabs/examples/demo.html

MaelthiS:
If I'm correct, shouldn't this fix be replicated to the Svelte, React, Vue packages ?

MaelthiS:
[...]

PR #1318, opened 2022-12-20
feat(@vtmn/react): keep the last focused VtmnTabsItem in a focus state

MaelthiS:
Original issue : https://github.com/Decathlon/vitamin-web/issues/1265

* When we focus on a VtmnTabsItem, and then focus outside the component, we lose the attribute aria-selected=true on the last focused VtmnTabsItem. This makes strange behavior because at least one of the VtmnTabsItem should always have the focus state.

Example here https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html. You can see that one tab always have a focused state, or a visual indication that it was the last consulted.

## Context
This change is required because this should be the default behavior of a Tab system, one of the tab should always have an indication that it is active. Tabs always have one tab opened, or active. 

- No


lauthieb (commented):
Hi @MaelthiS,

Thanks for your proposition.

I agree with what you said. But I have a problem regarding the responsibility of the `VtmnTabsItem`, I think this check needs to be done on `VtmnTabs` which is responsible to manage all items and not did inside each item (because it's only concerned is the last). Do you think it could be possible to propose a change at `VtmnTabs` level and spread down to the `VtmnTabsItems` concerned this update?

Also, can you please ensure that we will not have breaking changes for people that already consumes our `@vtmn/react` library?

Thank you very much for your help on this community library.

MaelthiS:
@lauthieb as today, the VtmnTabs calls just children, so we can't pass props in an easy way. This would need a rewrite of both components, and would add some breaking changes.
I can propose you to do this evolution in a second time, but as for now, we really need this fix. This could be the first step, and then we would do the evolution. What do you think ?

thibault-mahe:
Hi @MaelthiS, thanks for your work!
If I understand well, the onBlur doesn't seems to work that much, and your are suggesting to remove all aria-selected of the sibling on each focus, to only set the aria-selected on the active one.

I'll join Laurent on this one, do you think there would be a way to manage the state of the component higher up on the parent (like a compound component for example) and possibly also to get out of this imperative approach?

Thanks! :)

thibault-mahe:
Hello again @MaelthiS,

after discussion with @lauthieb, and because of your emergency, we will validate this PR as-is so that you can move forward. And then do as you said, i.e. rework with you the component to make it cleaner.

Before any merge, and this is a question that is systematically asked in the PR, can you make sure that these modifications will not bring any breaking changes for those who already use the component please?

Thank you!

lauthieb (approved):

thibault-mahe:
Hello @MaelthiS, is this PR still relevant? Have you been able to determine if it brought any breaking changes?
Thanks :)

lauthieb:
Hello @MaelthiS, thanks for this!
So for me it's ok. I let @thibault-mahe merge :)

thibault-mahe:
Done, thanks again @MaelthiS :)

PR #1321, opened 2022-12-29
feat(@vtmn/css): add href property on `VtmnListItem`

Tlahey:
## Changes description

Add a link wrapper on the VtmnListItem in order to redirect the user to a page.
Add new properties : 
- href
- rel
- target (default _blank)

## Context

We have to develop a list of suggestion link for the search on dkt website.
Today the only way is to use the event and javascript to redirect the user.
The solution is add a `<a>` around the content of the link item to enabled a redirection on a list item.

## Checklist

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch**. Please, don't request directly from your main!
- [ ] Check commits & PR names matches our requested structure. It must follow the https://www.conventionalcommits.org pattern.
- [ ] Check your code additions will fail neither code linting checks.
- [ ] I have reviewed the submitted code.
- [ ] If it includes design changes, please ask for a review with a core team designer.

No

## Other information

image
image


Tlahey:
PR in progress but it works on svelte storybook - part VtmnList.
Let's discuss and make the rest if everything is good :)

packages/sources/svelte/src/utils/link.js
```diff
@@ -0,0 +1,13 @@
...
+ * @param target
+ * @param rel
+ * @returns {string|*}
+ */
+export const computeRel = (target, rel) =>
+  target === '_blank'
+    ? Array.from(new Set(rel).add('noopener').add('noreferrer'))
+        .join(' ')
+        .trim()
+    : rel;
```
thibault-mahe, line 13: Could this be done with Typescript please ? I can't figure out here if `$$restProps['rel']` is returning a string, an array, or an object.

If `rel` is an object, then is not iterable and `new Set(rel)` sould not work.

If `rel` is a string, as it is in other framework, then`new Set(rel)` creates a set of the letters in the string:
```
Array.from(new Set('noopener'))
// ["n", "o", "p", "e", "r"]
```

then using the current script : 
```
computeRel('_blank', 'noopener nofollow')
// "n o p e r   f l w noopener noreferrer"
```

If `rel` is an array, then everything is fine :) (But maybe the arguments could be renamed or typed since there are intuitivelly considered as strings) 

Just FYI, we did something similar here, but with a split method as we were working with strings : https://github.com/Decathlon/vitamin-web/blob/main/packages/sources/react/src/components/actions/VtmnLink/VtmnLink.tsx#L50-L59

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
```diff
@@ -63,25 +83,57 @@
...
+          <slot name="subtext" />
+        {/if}
+      </div>
+    {/if}
+
+    {#if checkSlotExists('end-action')}
+      <div class="vtmn-list_end-action">
+        <slot name="end-action" />
+      </div>
+    {/if}
```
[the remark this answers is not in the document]
Tlahey: Not possible with svelte. 
The only way to avoid this is create another component with 

```
  {#if checkSlotExists('start-visual')}
    <div class="vtmn-list_start-visual">
      <slot name="start-visual" />
    </div>
  {/if}

  {#if checkSlotExists('text')}
    <div class="vtmn-list_text">
      <slot name="text" />

      {#if checkSlotExists('subtext')}
        <slot name="subtext" />
      {/if}
    </div>
  {/if}

  {#if checkSlotExists('end-action')}
    <div class="vtmn-list_end-action">
      <slot name="end-action" />
    </div>
  {/if}
```

And inject it link 

```
<li
  on:click
  class={componentClass}
  role="option"
  tabindex="0"
  aria-disabled={disabled}
  {...$$restProps}
>
  {#if href}
    <a
      class="vtmn-list__link"
      {href}
      {target}
      rel={computeRel(target, rel)}
      aria-disabled={disabled}
    >
    <ListContent>
      <slot name="start-visual" />
      <slot name="text" />
      <slot name="subtext" />
      <slot name="end-action" />
    </ListContent>
  </a>
{:else}
    <ListContent>
      <slot name="start-visual" />
      <slot name="text" />
      <slot name="subtext" />
      <slot name="end-action" />
    </ListContent>
{/else}
</li>
```

packages/sources/svelte/src/utils/link.js
Tlahey: Nice catch ! It should be 

```
new Set(rel && rel.split(' '))
```

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
```diff
@@ -63,25 +83,57 @@
...
-      <slot name="start-visual" />
-    </div>
-  {/if}
+  {#if href}
+    <a
+      class="vtmn-list__link"
+      {href}
+      {target}
+      rel={computeRel(target, rel)}
+      aria-disabled={disabled}
```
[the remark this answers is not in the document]
Tlahey: humm I don't know, the good practice of the list should at least add a item with a text. ?

packages/sources/css/src/components/structure/list/src/index.css
```diff
@@ -18,6 +18,21 @@
...
+  padding-right: 2.5rem; /* to avoid the padding-inline from child */
+}
+
+.vtmn-list li > a.vtmn-list__link[aria-disabled='true'] {
+  color: currentcolor;
+  display: inline-block; /* For IE11/ MS Edge bug */
+  pointer-events: none;
+  text-decoration: none;
+}
+
```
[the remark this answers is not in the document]
Tlahey: Today the button exist. It is the component without the href and you can access on the event with the on:click.

With this attribute, we want to add the redirection. So only the `<a>` will be affected by this css.

So yes and no, that because:
- the `.vtmn-list__link[aria-disabled='true']` is specific to the node `<a>`
- We can changes tomorrow with something else but it should be always the same behaviour on the top level (`<a>`). 

So keep the `<a>` selector for me is a secure option that said, if you put something else that a `<a>` on top, it will doesn't work.

packages/sources/css/src/components/structure/list/src/index.css
```diff
@@ -18,6 +18,21 @@
   padding: 0;
 }
 
+.vtmn-list a.vtmn-list__link {
+  display: flex;
+  inline-size: 100%;
+  text-decoration: inherit;
+  color: inherit;
+  padding-right: 2.5rem; /* to avoid the padding-inline from child */
```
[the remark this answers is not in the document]
crotru: also, size values of this file are written in pixel and get transpiled in rem in the end, so they should be written as `rem(40px)`.

This value might need change depending on `vtmn-list_item-size--...`

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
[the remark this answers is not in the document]
thibault-mahe: The list guidelines says that "Each item is composed of text with optional icons or images." indeed

packages/sources/css/src/components/structure/list/src/index.css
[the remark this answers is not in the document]
thibault-mahe: Not sure about the `It is the component without the href and you can access on the event with the on:click`. Should we not use an accessible and semantic button as we are doing for the `<a>` ?

packages/sources/css/src/components/structure/list/src/index.css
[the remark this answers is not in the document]
Tlahey: Will be remove, it was a workaround for an issue with the `display: inline-block;` (removed also)

packages/showcases/svelte/stories/components/structure/VtmnListItem/VtmnListItem.stories.svelte
```diff
@@ -19,6 +19,18 @@
...
+      control: { type: 'text' },
+    },
+    target: {
+      description: 'Target of the link.',
+      control: { type: 'text' },
+    },
+    rel: {
+      description: 'Rel of the link.',
+      control: { type: 'text' },
+    },
```
manglard, line 33: Maybe we can an example of the case we need, which is with a link AND a button as we can see in your screen shot. 

I don't know if it's a good practice but it can avoid some dirty usage of the component.
If the component allow it, we should show the best way to do it

packages/sources/vue/src/components/structure/VtmnListItem/VtmnListItem.vue
```diff
@@ -19,6 +19,15 @@ export default /*#__PURE__*/ defineComponent({
       type: Boolean as PropType<boolean>,
       default: false,
     },
+    href: {
+      type: String as PropType<string>,
+    },
+    rel: {
```
manglard, line 25: Why don't you use the `computeRel()` function ? Just to know

packages/showcases/vue/stories/components/structure/VtmnList/VtmnList.stories.js
```diff
@@ -7,22 +7,22 @@ export default {
...
         <template v-slot:text><span>Lorem ipsum</span></template>
         <template v-slot:subtext><span>Lorem ipsum dolor sit amet</span></template>
       </VtmnListItem>
-      <VtmnListItem>
+      <VtmnListItem href="/">
         <template v-slot:text><span>Lorem ipsum</span></template>
         <template v-slot:subtext><span>Lorem ipsum dolor sit amet</span></template>
       </VtmnListItem>
-      <VtmnListItem>
+      <VtmnListItem href="/">
```
manglard, line 25: Missing example with target as you have done for svelte

packages/showcases/react/stories/components/structure/VtmnList/VtmnList.stories.tsx
```diff
@@ -108,3 +110,20 @@ const SmallSizeTemplate: Story = (args) => (
 );
 
 export const SmallSize = SmallSizeTemplate.bind({});
+
+const WithLinkListTemplate: Story = (args) => (
+  <div style={{ width: '500px' }}>
+    <VtmnList {...args}>
+      {[...Array(3)].map((_, i) => (
+        <VtmnListItem key={i} size="small" href="/">
```
manglard, line 118: Same here maybe add an example with target

packages/sources/react/src/components/structure/VtmnList/VtmnList.tsx
```diff
@@ -60,12 +89,33 @@ export interface VtmnListItemProps
...
+   * Target of the link
+   * @type {string}
+   */
+  target?: string;
+
+  /**
+   * Rel of the link
+   * @type {string}
+   */
+  rel?: string;
```
manglard, line 109: Same here maybe we can use `computeRel` to be sure to have the same behavior no matter the language

packages/showcases/css/stories/components/structure/list/examples/small.html
```diff
@@ -136,3 +136,40 @@
...
+
+<div class="block">
+  <ul class="vtmn-list">
+    <li
+      class="vtmn-list_item-size--small vtmn-list_item--without-divider"
+      tabindex="0"
+    >
+      <a href="/">
+        <div class="vtmn-list_text">
+          <span>List Item without divider</span>
```
thibault-mahe, line 148: Are `span` mandatory for the style here? I have the impression that they don't bring anything else, no?

packages/sources/react/src/components/structure/VtmnList/VtmnList.tsx
```diff
@@ -33,6 +33,35 @@ export const VtmnListItemEndAction = ({
   </div>
 );
 
+/**
+ * Custom function to inject elements under the link
```
thibault-mahe, line 37: `Custom component` is more appropriate I think

packages/sources/react/src/components/structure/VtmnList/VtmnList.tsx
```diff
@@ -33,6 +33,35 @@ export const VtmnListItemEndAction = ({
...
+          React.isValidElement(child) && child.type !== VtmnListItemEndAction,
+      )}
+    </a>
+    {React.Children.toArray(children).filter(
+      (child) =>
+        React.isValidElement(child) && child.type === VtmnListItemEndAction,
+    )}
+  </>
+);
+
```
thibault-mahe, line 64: Could it be necessary to add an early return with : 
```
if (!React.Children) return <></>
```
?
What happens here if there is no React.Children ?

packages/sources/svelte/src/utils/link.js
thibault-mahe: This thread is marked as resolve but not fixed yet, is it ?

packages/sources/react/src/components/structure/VtmnList/VtmnList.tsx
thibault-mahe: Indeed +1

thibault-mahe (changes requested):

packages/sources/svelte/src/utils/link.js
Tlahey: Not fixed yet, I will do that with unit tests :)

packages/sources/react/src/components/structure/VtmnList/VtmnList.tsx
Tlahey: This behaviour doesn't exist on other langages.
In fact, the component should not worked if we don't put the text slot (VtmnListItemText)

packages/showcases/svelte/stories/components/structure/VtmnListItem/VtmnListItem.stories.svelte
Tlahey: Yes, we can do this easily with conditional controls -> https://storybook.js.org/docs/react/essentials/controls#conditional-controls

But in fact, this one doesn't work with svelte =/

manglard: you can't use <slot> in storybook ?

packages/showcases/css/stories/components/structure/list/examples/small.html
Tlahey: yes it's mandatory, the span are need if the listItem are secondary text under
image

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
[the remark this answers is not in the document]
Tlahey: Ok but aria-label will be the same as the `<a>`

packages/showcases/css/stories/components/structure/list/examples/small.html
thibault-mahe: ok :/

thibault-mahe: ok :/

thibault-mahe (dismissed):

packages/sources/css/src/components/structure/list/src/index.css
```diff
@@ -18,6 +18,19 @@
...
 }
 
+.vtmn-list a.vtmn-list__link {
+  display: flex;
+  inline-size: 100%;
+  text-decoration: inherit;
+  color: inherit;
+}
+
+.vtmn-list li > a.vtmn-list__link[aria-disabled='true'] {
```
manglard, line 28: Not appropriate to use `aria-disabled` maybe using `disabled` is more reliable

packages/sources/react/src/utils/link.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
manglard, line 1: Just to know is it possible to totally externalize this file to have only one to maintain ?

packages/sources/vue/src/components/structure/VtmnListItem/VtmnListItem.vue
```diff
@@ -40,16 +51,38 @@ export default /*#__PURE__*/ defineComponent({
...
-    <div v-if="$slots['start-visual']" class="vtmn-list_start-visual">
-      <slot name="start-visual" />
-    </div>
+    <template v-if="href">
+      <a
+        class="vtmn-list__link"
+        :href="href"
+        :target="target"
+        :rel="computeRel(target, rel)"
+        :aria-disabled="disabled"
```
manglard, line 60: Same here should have a way to set an `aria-label`

packages/sources/css/src/components/structure/list/src/index.css
Tlahey: Not possible -> https://www.scottohara.me/blog/2021/05/28/disabled-links.html
The best way here is to used `aria-disabled` on the link

packages/sources/react/src/utils/link.ts
Tlahey: Not possible today, the file are different between svelte / other libs (typescript).
The only way is to create an `util` lib.
But discuss with @lauthieb and @thibault-mahe we keep this file duplicate =/

packages/sources/vue/src/components/structure/VtmnListItem/VtmnListItem.vue
Tlahey: Same as above

packages/sources/css/src/components/structure/list/src/index.css
manglard: Well, can we create a status class instead ? Like `is-disabled`?

manglard: @thibault-mahe what's your point of view ?

thibault-mahe: `is-disabled` would be for the style, but `aria-disabled` is the real thing for assistive technology. Why is it not appropriate ?

manglard: because we base appearence of the button on the assertive attribute while styles is for everyone.
For me, aria is a complement to help screen readers to better understand our interfaces but not to use it as a CSS Selector.

As we discuss with Antoine, for this particular case, it's a good solution, but we missing some rules/good practices about the fact to use aria attributes as css selectors.

thibault-mahe: Yes you right, I didn't fully understand the context, my bad.
The CSS must not have any dependency with the DOM structure, I totally agree.
Currently our css is already very dependent on the DOM so this is something we will work on for v1

manglard (approved):

PR #1325, opened 2023-01-06
fix(@vtmn/css-toggle): use gap instead of margin

romuleald:
## Changes description
Minor change to use gap instead of margin

## Context
Allow to use easily reflex reverse to flip the label and the input

- No

### DIFF
--- a/packages/sources/css/src/components/selection-controls/toggle/src/index.css
+++ b/packages/sources/css/src/components/selection-controls/toggle/src/index.css
@@ -9,6 +9,7 @@
 .vtmn-toggle {
   display: flex;
   align-items: center;
+  gap: rem(8px);
 }
 
 .vtmn-toggle_switch {
@@ -118,7 +119,6 @@
   font-family: var(--vtmn-typo_font-family);
   color: var(--vtmn-semantic-color_content-primary);
   font-size: var(--vtmn-typo_text-2-font-size);
-  margin-inline-start: rem(12px);
 }
 
 /* Disabled state */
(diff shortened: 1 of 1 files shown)


packages/sources/css/src/components/selection-controls/toggle/src/index.css
```diff
@@ -118,7 +119,6 @@
   font-family: var(--vtmn-typo_font-family);
   color: var(--vtmn-semantic-color_content-primary);
   font-size: var(--vtmn-typo_text-2-font-size);
-  margin-inline-start: rem(12px);
```
romuleald, line 121: **question**: I removed this one since it seemed to me it was the same as the medium, but maybe I'm missing a point here!

thibault-mahe:
Hello @romuleald, thanks for the suggestion!

I understand that `gap` can be a more logical property than `margin-inline-start`, no disagreement on that point.

However, from what I understand, the component now has a gap of 0.5rem (`rem(8px)`) regardless of its size. The initial spacing in `--small` is also 0.5 ; on the other hand as you point out in your comment the spacing on `--medium` / by default is 0.75rem (`rem(12px)`).

Shouldn't we either propose a gap for each toggle size, or default it to `rem(12px)` and modify this gap for the small size ?

Thank you!

lauthieb:
Hello,
I'm closing this Pull Request because there is no more activity for several weeks.
Feel free to reopen it whenever you want to take up this topic.
Thanks again for your contribution. :)
Laurent

PR #1334, opened 2023-01-18
fix(@vtmn/svelte): fix events accessibility `VtmnListItem`

Tlahey:
With the last version of `VtmnListItem` we have found some issues with accessibility.

This PR fixed all issue due to keypress / click / stop propagation on children.
For example : 
- `<li>` click action are not possible with keyboard, only mouse.

No breaking changes, just changes on the role by default for `VtmnListItem`

### DIFF
--- a/packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
+++ b/packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
@@ -3,7 +3,7 @@
   import { cn } from '../../../utils/classnames';
   import { computeRel } from '../../../utils/link';
   import { createEventDispatcher } from 'svelte';
-  import { ENTER, SPACE } from '../../../utils/keyCodes';
+  import { ENTER } from '../../../utils/keyCodes';
 
   /** @restProps */
 
--- a/packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
+++ b/packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
@@ -9,7 +9,6 @@
 
   const SLOTS = $$props.$$slots;
   const dispatch = createEventDispatcher();
-  let listElement;
 
   /**
    * The size of the list.
@@ -26,6 +25,13 @@
    * @defaultValue true
    */
   export let divider = true;
+
+  /**
+   * role of the list element
+   * @type {string}
+   * @default listitem
+   */
+  export let role = 'listitem';
 
   /**
    * Set disabled state of list item.
@@ -86,11 +92,10 @@
 <!-- svelte-ignore a11y-role-has-required-aria-props -->
 <!-- This will be refactored in next major release -->
 <li
-  bind:this={listElement}
-  on:click|stopPropagation={handleSelectItem}
-  on:keydown|stopPropagation={handleSelectItem}
+  on:click={handleSelectItem}
+  on:keydown={handleSelectItem}
   class={componentClass}
-  role="option"
+  {role}
   tabindex={href ? -1 : 0}
   aria-disabled={disabled}
   {...$$restProps}
@@ -108,11 +113,7 @@
       aria-disabled={disabled}
     >
       {#if checkSlotExists('start-visual')}
-        <div
-          class="vtmn-list_start-visual"
-          on:click={(e) => e.stopPropagation()}
-          on:keydown={(e) => e.stopPropagation()}
-        >
+        <div class="vtmn-list_start-visual">
           <slot name="start-visual" />
         </div>
       {/if}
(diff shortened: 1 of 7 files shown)


thibault-mahe:
This PR seems to bring some fix to allow nested interactive elements. We should take extra care with this kind of decision since it could confuse assistive technologies and hurt accessibility.

See https://dequeuniversity.com/rules/axe/4.5/nested-interactive or https://adrianroselli.com/2016/12/be-wary-of-nesting-roles.html for instance.

@manglard @gaspard-mathon what are your opinion about that ?

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
```diff
@@ -71,12 +75,20 @@
...
 
 <!-- svelte-ignore a11y-role-has-required-aria-props -->
 <!-- This will be refactored in next major release -->
 <li
-  on:click
+  bind:this={listElement}
+  on:click|stopPropagation={handleSelectItem}
+  on:keydown|stopPropagation={handleSelectItem}
   class={componentClass}
   role="option"
```
manglard, line 93: If element is clickable maybe a `role="button"` is more appropriate

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
```diff
@@ -89,12 +101,18 @@
...
+      on:keydown={(e) => e.stopPropagation()}
       rel={computeRel(target, rel)}
       aria-label={$$restProps['aria-label']}
       aria-disabled={disabled}
     >
       {#if checkSlotExists('start-visual')}
-        <div class="vtmn-list_start-visual">
+        <div
+          class="vtmn-list_start-visual"
+          on:click={(e) => e.stopPropagation()}
```
manglard, line 113: Do you try to a `pointer-event:none` instead stop propagation ?

packages/sources/svelte/src/utils/keyCodes.js
```diff
@@ -0,0 +1,10 @@
+export const ENTER = 13;
```
manglard, line 1: is it compatible for all countries ? like china ?

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
Tlahey: yes, by default the css will displayed an "option" rôle. 
Should we change it on button for every `VtmnListItem` ?

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
Tlahey: No, we wants to stop the propagation of parent click event here by default, and just trigger the click on the child.
`point-event: none` will disable every children action.

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
manglard: I guess for screen reader it will be clearer if it's identified like a button. You should try because i'm not sure.

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
manglard: maybe we can discuss that because i'm not sure to understand well

lauthieb (dismissed):

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
Tlahey: This issue are resolved, we don't apply propagation on the `VtmnListItem`.
Because of the behaviour of the component, we have to set manually the `stopPropagation` on the slot's element.

packages/sources/svelte/src/components/structure/VtmnListItem/VtmnListItem.svelte
```diff
@@ -71,12 +75,20 @@
...
+      dispatch('click');
+    }
+  };
 </script>
 
 <!-- svelte-ignore a11y-role-has-required-aria-props -->
 <!-- This will be refactored in next major release -->
 <li
-  on:click
+  bind:this={listElement}
```
Tlahey, line 89: can be removed

packages/sources/svelte/src/components/structure/VtmnListItem/test/VtmnListItem.spec.js
```diff
@@ -61,6 +63,24 @@ describe('VtmnListItem', () => {
...
+    expect(handleClick).toHaveBeenCalledTimes(1);
+  });
+  test('Should trigger once click if href + click slot button', async () => {
+    const { getByRole, component } = render(VtmnListItem, {
+      href: 'https://decathlon.fr',
+    });
+    const handleClick = jest.fn();
+    component.$on('click', handleClick);
+    await fireEvent.click(getByRole('button'));
+    // expect(handleClick).toHaveBeenCalledTimes(1);
```
Tlahey, line 82: fix the test

thibault-mahe (approved):

Issue #1322, opened 2023-01-03
bug: failed to parse source map on project using React with Typescript and Vtmn Library for React


jdeca-decat:
**Describe the bug**
When I use "npm run build" on my project using React Typescript and Vtmn Library for React, I've got warnings (view image) : "failed to parse source map..."
![vtmn-problem-build](https://user-images.githubusercontent.com/107050404/210386909-b082a68f-bc40-4695-805e-c2bee51a65f3.png)
A solution consist to add "GENERATE_SOURCEMAP=false" to temporarily avoid this problem, to the command line "build" in the package.json, but I would like to find a cleaner solution because "today's warnings will be tomorrow's errors" !

**Steps to reproduce**
    1. Create a React Typescript project: "npx create-react-app [nameApp] --template typescript"
    2. Install Vtmn Library for React "npm install @vtmn/react"
    3. Create a simple page with a Vtmn Component
    4. Build your code and see warnings

**Version affected**
@react: 18.2.0
@typescript: 4.9.4
@vtmn/react: 0.57.0
@react-scripts: 5.0.1
@eslint: 8.31.0

jdeca-decat:
It seem that the error does not come from vitamin-web -> (https://github.com/facebook/create-react-app/pull/11752)
Waiting for a resolution from Facebook...

thibault-mahe:
Hi @jdeca-decat ! Thanks for reporting this issue.
I was able to replicate the concern on my side indeed.

I don't think the problem is with CRA, the PR you mentioned only asks for the possibility to override the warnings. But if the warnings are present it's probably for a good reason, or we shouldn't use source maps (to be discussed).

The issue that the warnings indicate here is that our packages distribute maps that reference sources that are not present. And indeed we don't export our sources in our packages, so the mapping can't work. To use your exemple : 

- `./node_modules/@vtmn/react/dist/index.js` uses the distributed `./node_modules/@vtmn/react/dist/index.js.map` to map between the production files and the original Typescript source files.
- `index.js.map` try to map a file to `node_modules/@vtmn/react/src/components/overlays/VtmnTooltip/VtmnTooltip.tsx` but we do not distribute original `@vtmn/react/src` in our package, so the reference is missing
- Warning: Failed to parse source map / no such file or directory / etc.

[...]

PR #1362, opened 2023-02-08
build(@vtmn/react): remove sourcemap and improve treeshaking

Shyrro:
Signed-off-by: Shyrro <zakaria.sahmane@decathlon.com>

## Changes description

Removes sourcemap to solve this issue : https://github.com/Decathlon/vitamin-web/issues/1322

Since sourcemaps are mainly for debug purposes, we're also splitting the files to improve treeshake and thus improve debugging capabilities.

## Context
It solves two problems : 

- Sourcemaps warnings 
- Improves bundle size ( the fact that sourcemaps are gone makes the package around 200kb lighter, and the fact that it is now treeshakable also makes consumer apps lighter )

## Checklist

- No

### DIFF
--- a/packages/sources/react/package.json
+++ b/packages/sources/react/package.json
@@ -19,9 +19,10 @@
   },
   "sideEffects": false,
   "license": "Apache-2.0",
-  "main": "dist/index.js",
-  "module": "dist/index.es.js",
-  "jsnext:main": "dist/index.es.js",
+  "type": "module",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
+  "jsnext:main": "dist/index.js",
   "types": "dist/src/index.d.ts",
   "files": [
     "dist"
--- a/packages/sources/react/package.json
+++ b/packages/sources/react/package.json
@@ -17,7 +17,6 @@
     "type": "git",
     "url": "git+https://github.com/Decathlon/vitamin-web.git"
   },
-  "sideEffects": false,
   "license": "Apache-2.0",
   "type": "module",
   "main": "dist/index.cjs",
--- a/packages/sources/react/rollup.config.js
+++ b/packages/sources/react/rollup.config.js
@@ -5,24 +5,22 @@ import url from '@rollup/plugin-url';
 import svgr from '@svgr/rollup';
 import external from 'rollup-plugin-peer-deps-external';
 import postcss from 'rollup-plugin-postcss';
-import pkg from './package.json';
 
-const TARGETS = [
-  { format: 'cjs', dir: './dist', tsconfig: './tsconfig.cjs.json' },
-  { format: 'es', file: pkg.module, tsconfig: './tsconfig.es.json' },
-];
+const formatOutput = (format) => ({
+  format,
+  exports: 'named',
+  sourcemap: false,
+  dir: 'dist',
+  preserveModules: true,
+  preserveModulesRoot: 'src',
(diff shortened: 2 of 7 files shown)


packages/sources/react/package.json
```diff
@@ -19,9 +19,10 @@
   },
   "sideEffects": false,
   "license": "Apache-2.0",
-  "main": "dist/index.js",
-  "module": "dist/index.es.js",
-  "jsnext:main": "dist/index.es.js",
+  "type": "module",
```
thibault-mahe, line 22: @Shyrro could this be a breaking change for consumers using node 14 and/or the require syntax? Do we have a mean to know if we have such users?

packages/sources/react/rollup.config.js
```diff
@@ -5,24 +5,22 @@ import url from '@rollup/plugin-url';
 import svgr from '@svgr/rollup';
 import external from 'rollup-plugin-peer-deps-external';
 import postcss from 'rollup-plugin-postcss';
-import pkg from './package.json';
 
-const TARGETS = [
-  { format: 'cjs', dir: './dist', tsconfig: './tsconfig.cjs.json' },
-  { format: 'es', file: pkg.module, tsconfig: './tsconfig.es.json' },
-];
```
thibault-mahe, line 13: @Shyrro I don't know the previous build very well, but there is no need for that granularity here? i.e. to be able to differentiate a different dist, a different config, etc.

I understand the use of ternary below, but aren't there cases where we would potentially want to export something other than es module or common js?

packages/sources/react/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "noFallthroughCasesInSwitch": true,
     "rootDir": ".",
     "declaration": true,
-    "sourceMap": true,
+    "declarationDir": "dist",
```
thibault-mahe, line 9: @Shyrro i'm a bit curious on this change, is this more "performant" to precise that the declarations are in the dist?

Shyrro: @thibault-mahe It is not more performant, but rather to insure that the declarations will always go to the dist, by default they're put in the working directory, so i'd rather limit the confusion

packages/sources/react/rollup.config.js
Shyrro: @thibault-mahe Actually, we can do the same and have as much outputs as we wants without this. 

The main difference here is that before, since we used the `TARGETS.map` as export default for our config, this meant that rollup used to run the build command twice with each configuration. Whereas now, we only run the build command once and have the right files exported

packages/sources/react/package.json
Shyrro: @thibault-mahe Uhm, i don't know if we have such users, but the `type: module` just indicated that all the files that end with `.js` are `esmodules` by default, if someone uses `require`, the target becomes the main and uses the `cjs` export instead. 
I don't think it breaks on node 14 but would be interested to test it

thibault-mahe (dismissed):

thibault-mahe:
Thanks for your quick answer @Shyrro , and thanks for this contribution!
I'm merging it :)

lauthieb (dismissed):

lauthieb (approved):

jdeca-decat:
@thibault-mahe @Shyrro @lauthieb Hello, thank you very much for fixing the issue #1322, I have no more warnings now :)

PR #1364, opened 2023-02-08
feat: make extensible `Quantity` component

### LINKED ISSUE #1356: feat: make extensible `Quantity` component
**Is your feature request related to a problem? Please describe.**

Actually, QuantityPicker has a fixed size.

**Describe the solution you'd like**

But depending on the content we need to make it extensible.

Related issue: https://github.com/Decathlon/vitamin-design/issues/162

lauthieb:
## Changes description

Closes #1356 

As we need to make it extensible without making breaking changes, I propose to add a new prop called `fullWidth` (boolean by default === 'false').

CleanShot 2023-02-08 at 21 40 39@2x
CleanShot 2023-02-08 at 21 52 08@2x
CleanShot 2023-02-08 at 22 10 00@2x
CleanShot 2023-02-08 at 21 52 27@2x

For your information, in Figma, there's no prop (variant) because the behavior is like this: 
CleanShot 2023-02-08 at 22 14 34. That was I think the best thing to reproduce this behavior is to set a full width based on the quantity's parent.

## Checklist

- No

### DIFF
--- a/packages/showcases/core/csf/components/selection-controls/quantity.csf.js
+++ b/packages/showcases/core/csf/components/selection-controls/quantity.csf.js
@@ -49,6 +49,14 @@ export const argTypes = {
       type: 'boolean',
     },
   },
+  fullWidth: {
+    type: { name: 'boolean', required: false },
+    description: `Displays the quantity at 100% of its parent's width`,
+    defaultValue: false,
+    control: {
+      type: 'boolean',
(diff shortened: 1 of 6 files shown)


thibault-mahe (approved):
Thank you for this addition and for the very clear and illustrated presentation of the PR. An example to follow!

PR #1378, opened 2023-02-25
fix(@vtmn/svelte, @vtmn/react, @vtmn/vue): fix rounded value on `VtmnRating`

Tlahey:
## Changes description

After checking the Decathlon website, I saw many products that the stars are stuck at 4.5.
image

But some of them are a rating > 4.7.

## Context

The rule on the current component is : 
- If my value are a floating value, always round on the 0.5 above.

But here with a value > 4.5 we don't round.

The question here is, what we defined when a number is > 4.5.
I've discuss with @lauthieb, we need to speak about it, maybe take the example of Google that round the value with the nearest one.

And write this rule on the Design System website so that there are no more questions about it.

## Checklist

No breaking changes

## Other information

This PR aims to enhancer the storybook
- Add conditional values in order to display only specific fields when `readonly` are true / false

| Readonly = false |  Readonly = true |
|---|---|
|image|image |

Improve tests
- Move the function who compute the `VtmnIcon` who has to be displayed. Add more tests on this function.
- Add more test on the component

### DIFF
--- a/packages/showcases/core/csf/components/indicators/rating.csf.js
+++ b/packages/showcases/core/csf/components/indicators/rating.csf.js
@@ -59,7 +59,7 @@ export const argTypes = {
     description:
       'Rating value. NB: if readonly is set to false, the value step is 1.',
     defaultValue: 0,
-    control: { type: 'range', min: 0, max: 5, step: 0.5 },
+    control: { type: 'range', min: 0, max: 5, step: 0.1 },
   },
   comments: {
     type: { name: 'string', required: false },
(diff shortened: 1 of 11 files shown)


packages/showcases/core/csf/components/indicators/rating.csf.js
```diff
@@ -51,10 +52,11 @@ export const argTypes = {
     description: 'Display compact mode. Only if readonly is true.',
     defaultValue: false,
     control: { type: 'boolean' },
+    if: { arg: 'readonly' },
   },
   value: {
     type: { name: 'number', required: false },
-    description: 'Rating value.',
+    description: 'Rating value. NB: if readonly are false, value step is 1.',
```
lauthieb, line 59:
```suggestion
    description: 'Rating value. NB: if readonly is set to false, the value step is 1.',
```

lauthieb (approved):

lauthieb:
Thanks @Tlahey, that's ok for me!
I also added it this fix for React & Vue.

PR #1388, opened 2023-03-13
refactor(vtmn/icons): reorder font src, woff2 in first position for performance

jquinc30:
## Changes description

Reorder src of vitamix `@font-face`

## Context

The browser download and use the first font file supported so the order is important to ensure best performance.
Before is was ttf, woff, woff2 so the browser use the ttf version (62K)
After it's woff2, woff, ttf so it will take the woff2 version which is better (26K, 59% lighter 🚀 )

NB: it's a progressive enhancement, for [oldest](https://caniuse.com/woff2) browsers that doesn't support woff2 nor woff, they will continue to use the ttf version as before

## Checklist

- No

### DIFF
--- a/packages/sources/icons/src/generated/vitamix/font/vitamix.css
+++ b/packages/sources/icons/src/generated/vitamix/font/vitamix.css
@@ -1,8 +1,8 @@
 @font-face {
     font-family: "vitamix";
-    src: url("./vitamix.ttf?8880a93e7dbb625d98d1ee364132f425") format("truetype"),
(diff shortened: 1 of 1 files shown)


lauthieb (approved):

lauthieb:
Thanks a lot @jquinc30 for your precious advices!

Tlahey:
Nice catch @jquinc30 :)

thibault-mahe (approved):
Nice catch indeed, thanks @jquinc30 :)

PR #1411, opened 2023-04-06
fix(@vtmn/svelte, @vtmn/vue, @vtmn/react): interactive `VtmnRating` have duplicate `$$restProps`

### LINKED ISSUE #1385: bug(@vtmn/svelte): component `VtmnRating` call many times `$$restProps` on child
image

There is an issue on the restProps part, here, if the user apply a class on the `VtmnRating` component, this one will be applied on the main node and the first child (only if it is not readonly).

We have to defined where the data must be applied. And if the second div should apply the restProps.

Tlahey:
Closes #1385 

## Changes description
Add on the `VtmnRating` svelte component a check on the restProps in order to not apply it on the root level of the component and under if the rating is interactive.
No changes on the class, only on props like aria-label etc ...

## Context
In some case, if the component is interactive, the class are applied twice, on the root level and under.

## Checklist

- No

Tested on svelte / react / vue on storybook.

### DIFF
--- a/packages/sources/react/src/components/indicators/VtmnRating/VtmnRating.tsx
+++ b/packages/sources/react/src/components/indicators/VtmnRating/VtmnRating.tsx
@@ -105,7 +105,7 @@ export const VtmnRating = ({
         className,
       )}
       aria-disabled={disabled}
-      {...props}
+      {...(readonly && props)}
     >
       {/**
        * Interactive mode
@@ -116,6 +116,7 @@ export const VtmnRating = ({
           aria-label="Rate the article"
           role="radiogroup"
           data-rating={position}
+          {...props}
         >
           {Array.from(Array(5).keys()).map((index) => (
             <React.Fragment key={`rating-${index + 1}`}>
--- a/packages/sources/svelte/src/components/indicators/VtmnRating/VtmnRating.svelte
+++ b/packages/sources/svelte/src/components/indicators/VtmnRating/VtmnRating.svelte
@@ -72,7 +72,11 @@
   $: starsCnt = compact && readonly ? 1 : 5;
 </script>
 
-<div class={componentClass} aria-disabled={disabled} {...$$restProps}>
+<div
+  class={componentClass}
+  aria-disabled={disabled}
+  {...readonly && $$restProps}
+>
   {#if readonly}
     {#each Array(starsCnt) as _, index}
       {@const position = index + 1}
--- a/packages/sources/svelte/src/components/indicators/VtmnRating/test/VtmnRating.spec.js
+++ b/packages/sources/svelte/src/components/indicators/VtmnRating/test/VtmnRating.spec.js
@@ -354,6 +354,16 @@ describe('VtmnRating', () => {
       expect(inputs[1].checked).toEqual(true);
       expect(getInteractive(container)).toHaveAttribute('data-rating', '2');
     });
+
+    test('Should not contains restProps many times', () => {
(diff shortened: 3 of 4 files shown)


lauthieb (approved):

lauthieb:
Thanks @Tlahey, that's ok for me. I'll let @thibault-mahe review next Tuesday.

packages/sources/react/src/components/indicators/VtmnRating/VtmnRating.tsx
```diff
@@ -105,7 +105,7 @@ export const VtmnRating = ({
         className,
       )}
       aria-disabled={disabled}
-      {...props}
+      {...(readonly && props)}
```
thibault-mahe, line 108: 2 questions:
- it is not clear to me why there is a condition on the readonly props?
- no tested but does this work? Since you are spreading, shouldn't you use a ternary here like `{...(readonly ? props : {})}` ?

packages/sources/svelte/src/components/indicators/VtmnRating/VtmnRating.svelte
```diff
@@ -72,7 +72,11 @@
   $: starsCnt = compact && readonly ? 1 : 5;
 </script>
 
-<div class={componentClass} aria-disabled={disabled} {...$$restProps}>
+<div
+  class={componentClass}
+  aria-disabled={disabled}
+  {...readonly && $$restProps}
```
thibault-mahe, line 78: same issue here + should you add parentheses here?

packages/sources/svelte/src/components/indicators/VtmnRating/test/VtmnRating.spec.js
```diff
@@ -354,6 +354,16 @@ describe('VtmnRating', () => {
       expect(inputs[1].checked).toEqual(true);
       expect(getInteractive(container)).toHaveAttribute('data-rating', '2');
     });
+
+    test('Should not contains restProps many times', () => {
```
thibault-mahe, line 358: `Should not have repeated restProps`

packages/sources/react/src/components/indicators/VtmnRating/VtmnRating.tsx
Tlahey: For the ternary, it is not necesary, 
image

The condition is on the readonly part because, if we are on readonly true / false, the structure are not the same. 

image

With readonly false, we have a sub level with `vtmn-rating--interactive` which is contains all properties like aria-labels etc ... who the user can override with the restProps. 

This level doesn't exist if readonly is true.
image

packages/sources/svelte/src/components/indicators/VtmnRating/VtmnRating.svelte
Tlahey: yes indeed

lauthieb (approved):

packages/sources/react/src/components/indicators/VtmnRating/VtmnRating.tsx
lauthieb: Thanks @Tlahey for this explanation.
Since @thibault-mahe is on vacation this week and is replacing him during his absence, this PR is open and your explanations seem to be correct. I will therefore take the decision to merge this PR. 
@thibault-mahe, don't hesitate to come back to it if you need to make a modification to these changes via a patch when you come back. 
Thanks!

PR #1458, opened 2023-08-02
fix: convert breakpoints variables to values instead of strings

AxelPeter:
## Changes description

Change value of Breakpoints design tokens from strings to values to be used in `var()`

## Context

[With Web Checkout](https://github.com/dktunited/web-checkout-front/pull/89), we try to rely the most on Vitamin design tokens. But i have an issue when using the breakpoints design tokens. Currently, they are strings and can't be used as value in `var()`

```css
/* Don't work */
--foo: "900px";
max-width: var(--foo);

/* Works */
--foo: 900px;
max-width: var(--foo);
```

## Checklist

If these tokens are used as strings somewhere, it could be a breaking change.
But any others design tokens are values and not strings. I've checked in the codebase any use as strings but didn't found any, so it should not be a problem here

### DIFF
--- a/packages/sources/css/src/design-tokens/src/breakpoints.css
+++ b/packages/sources/css/src/design-tokens/src/breakpoints.css
@@ -1,10 +1,10 @@
 :root {
-  --vtmn-breakpoint_mobile-max: '599px';
-  --vtmn-breakpoint_tablet-min: '600px';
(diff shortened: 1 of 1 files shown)


lauthieb:
Thanks @AxelPeter for this PR, good catch!
I just checked and I think this will not have consequences for our existing consumers. So we can merge :)

PR #1462, opened 2023-09-26
fix(@VtmnReact): allow badge component to have 0 as a value

romuleald:
## Changes description
Badge doesn't accept 0 as a value, but instead it's showing an empty string
image image

## Context
It should accept 0 as a value as fixed here
image

## Checklist

- [ ] If it includes design changes, please ask for a review `design-system-core-team-design` GitHub team.

- Yes
- No

### DIFF
--- a/packages/sources/react/src/components/indicators/VtmnBadge/VtmnBadge.tsx
+++ b/packages/sources/react/src/components/indicators/VtmnBadge/VtmnBadge.tsx
@@ -29,7 +29,7 @@ export const VtmnBadge = ({
       }`}
       {...props}
     >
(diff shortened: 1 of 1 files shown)


thibault-mahe:
Hi @romuleald, thank you for suggesting this PR.

We've checked with the designers and, as @Tlahey also points out in his comment, this choice is indeed a documented design decision.

Given current priorities and our resources, there are no plans to change this behavior on this version of Vitamin. On the other hand, we're taking good note of it so that we can take it into account in the new version we're currently working on.

I'm therefore closing this PR, but do note we're very grateful for the time you spent suggesting this code.

Thanks!

